PR #16, opened 2021-12-06
Delete one or multiple

tijlleenders:


tijlleenders:
Show a delete button only when there are children listed.

After pressing delete button children shake in fear of being deleted.
Touch/click a child to confirm delete and they go away forever, including all recursive (exclusive) descendants.
Touch the delete button again to turn off delete mode.

Issue #147, opened 2022-03-23
Refactor App.scss into multiple files


pulkitvyas08:
Just observed that `App.scss` is a 650+ LOC file containing styling code from almost all components. We need to segment it heavily before it causes more readability issues

PR #150, opened 2022-03-24
Add Dark theme CSS for 'My Feelings' page

mallikarai05:


src/App.css
```diff
@@ -345,7 +345,14 @@
```
pulkitvyas08, line 345: I highlighted in #147 that the CSS files are becoming too long and need to be segmented into multiple files. Approving for now but something to make a PR about for sure

pulkitvyas08 (approved):

PR #174, opened 2022-04-05
Fix logged in router 404 and resolve warnings

### LINKED ISSUE #172: Fix routing issue on prod
With #169 and the way `react-router` is setup, we now have this situation - 
- The default (non-logged) user gets on the landing page at https://zinzen.vercel.app/
- Logged users get `404` on https://zinzen.vercel.app/

pulkitvyas08:
Resolves #172


pulkitvyas08 (commented):
Working as intended, verified through the preview

Issue #148, opened 2022-03-23
Navbar buttons not working + remove menu


pulkitvyas08:
Seems like their `OnClick` has not been connected with the router yet.
Also the logo or the ZinZen link on the navbar should redirect to homepage.
Remove the stuff marked with red X.
![image](https://user-images.githubusercontent.com/44887755/162404409-4318fab4-0959-4713-8351-8ab6ca2a6d6b.png)

pulkitvyas08:
Also the logo on the navbar should redirect to homepage

tijlleenders:
On the choose theme page the menu button is still showing.

pulkitvyas08:
> On the choose theme page the menu button is still showing.

Fixed in #218

PR #190, opened 2022-04-07
Add ... when text length is too long

mallikarai05:
Feature: Whenever a text is too big for the button use "..." #94


src/components/dashboard/UserChoiceDashboard.js
```diff
@@ -16,6 +16,10 @@ export const UserChoiceDashboard = () => {
     const { t } = useTranslation();
     const darkModeStatus = useRecoilValue(darkModeState);
 
+    function truncateContent(content, maxLength = 20) {
+        return content.substring(0, maxLength) + '...'
```
pulkitvyas08, line 20: This is adding `...` to every text in a button. Need to change this to put `...`'s only when the `maxLength` is crossed. Please fix

pulkitvyas08 (changes requested):

src/components/dashboard/HeaderDashboard.js
```diff
@@ -42,7 +42,7 @@ export const HeaderDashboard = () => {
                 <Navbar.Collapse id="responsive-navbar-nav">
                     <Nav className="navbar-custom">
                         <Nav.Link
-                            href="#home"
+                            href="https://zinzen.vercel.app/Home"
```
pulkitvyas08, line 45: This seems like the code that would resolve #148. Please make sure to target only one issue per PR or at least mention explicitly when you are working on more than one issue within the same PR. I'm going to approve this for now but follow this convention moving forward

pulkitvyas08 (approved):
LGTM

PR #245, opened 2022-04-29
Style Update

ashiquemushtaq:
I have updated the CSS as asked.

Screenshot 2022-04-29 at 9 41 37 PM
Screenshot 2022-04-29 at 9 41 52 PM

### DIFF
--- a/src/components/dashboard/UserChoiceDashboard.jsx
+++ b/src/components/dashboard/UserChoiceDashboard.jsx
@@ -109,9 +109,9 @@ export const UserChoiceDashboard = () => {
                         size="lg"
                         className={
                             darkModeStatus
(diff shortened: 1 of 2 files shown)


tijlleenders:
Cool @ashiquemushtaq !

I see that the font color is not consistent.
In the light theme I expect it to be all dark, as so:
image

In the dark theme I expect it to be all light, as so:
image

We're designing for mobile first - so not an issue; but as a bonus can you check the + width on desktop?
This is the new version (too wide) : 
image

This is the old version:
image

ashiquemushtaq:
Okay i will look into that :)

ashiquemushtaq:
Made the font colors consistent.

PR #251, opened 2022-04-30
add vercel link as dev version

tijlleenders:


pulkitvyas08 (approved):
This was very much needed. Sometimes I would log in from my other accounts and cannot remember the Vercel link and even after opening the repository, there is no place where the dev link is mentioned apart from the deployment messages after merges. GTG from my side. Also adding an `environments` section to the repository will also make the dev deployment more visible

Issue #215, opened 2022-04-09
Add prettier + ESLint pre-commit hooks


tijlleenders:
[Example for Prettier + ESLint](https://gist.github.com/estorgio/e8bcaa8e87d0fcdcf85fdf598956e34c)

PR #267, opened 2022-05-05
Add Prettier and ESLint Pre-Commit Hooks

mallikarai05:
#215

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,12 +8,12 @@
     }
   },
   "lint-staged": {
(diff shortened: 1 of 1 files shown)


pulkitvyas08 (approved):
Looks alright from the implementation perspective but I have very little practical experience with pre-commit hooks so can't say it works. @mallikarai05 can you tell us if it fixes eslint issues on commits?

Issue #265, opened 2022-05-05
Feedback page should not refresh


tijlleenders:
Right now the feedback page refreshes.

You can see this clearly by opening developer tools (F12) and setting the network speed to 'Slow 3G' on the network tab.

![image](https://user-images.githubusercontent.com/44887755/166874429-07d5570c-81e1-4fb6-8d02-959b082d7aa3.png)

GithinjiHans:
Already handled it.

PR #269, opened 2022-05-05
Fix/refresh feedback 265

GithinjiHans:
Fixed the Feedback page from refreshing.


tijlleenders (approved):
Not sure if the yarn.lock should be in there but otherwise looking good!

Fixes #265

Issue #237, opened 2022-04-29
Convert the project to typescript and fix errors


tijlleenders:
Right now the project is React Javascript.

The project needs to be converted to typescript.
There's a [howto on converting **an existing project** here]( https://create-react-app.dev/docs/adding-typescript/).

pulkitvyas08:
Please note that some parts of this have already been implemented. For example, the typescript npm packages, as well as `tsconfig.json`, are already there in the project. Check them in `package.json` and project root. This was done as part of migrating the build system to `Vite` in #221 
What remains to be done is the following:
- [ ] Configuring the strictness of type checking within the project inside `tsconfig.json`
- [ ] Changing `.jsx` to `.tsx` and `.js` to `.ts`
- [ ] Adding interfaces and types
- [ ] Adding custom global interfaces (using `global.d.ts`) for packages that are not part of `@types/`

@Tushar-4781 maybe you would like to take point on this.

Please make sure to keep the following things in mind:
- `tsc --init` and other commands have already been executed and the relevant packages are there in the app. Modifications made to `tsconfig.json` due to executing those commands once again might cause breaking changes.
- Make frequent commits with small milestones as this is a big change to the overall project and will create conflicts for sure 
- Expanding on the above point, avoid using `@ts-ignore`, and wherever you don't see a suitable fix for a type check and are forced to use it, just make a commit and start a discussion in the PR itself.

I have some experience in migrating JS projects to TS, so I understand how frustrating the process could be in some cases. Hopefully, we will get this done ASAP

pulkitvyas08:
@tijlleenders the title of the issue says "and fix errors" is this referring to any particular error present in the project at the moment or about fixing the errors that would be cause by converting to TS?

tijlleenders:
Good question @pulkitvyas08 .  
Let's clarify: it is about fixing the errors caused by the conversion.

pulkitvyas08:
@Tushar-4781 please don't consider my work on #117 as a blocker for this task. I will implement this feature in TypeScript itself to make it easier for everyone involved

Tushar-4781:
Resolved #283

PR #283, opened 2022-05-06
Typescript integration

Tushar-4781:
This PR corresponds to #237. There are two problems I m facing: - 

1. There is a problem with the imports. Although I did try to add tsconfigpaths in tsconfig.json still it's showing a problem. Below is the snapshot. This problem is common for all the imports
image

2. And another one is this. Like how do I define the type for this RecoilState 
image


Tushar-4781:
Kindly ignore the convention problems. It's not detected by the eslint. I have to check that part. Help me with the above 2 problems

src/Interfaces/ILanguage.ts
```diff
@@ -0,0 +1,9 @@
+export interface ILanguage {
+    sno: Number;
+    title: string;
+    langId: string;
+}
+
+export interface ILanguageListProps{
```
pulkitvyas08, line 7: Add a space between `ILanguageListProps` and `{`

src/components/languagechoices/LangItem.tsx
```diff
@@ -1,25 +1,25 @@
-/* eslint-disable react/prop-types */
 import React from 'react';
 import { Button } from 'react-bootstrap';
 import { useRecoilState } from 'recoil';
 import i18n from 'i18next';
 
+import { ILanguage } from 'Interfaces/ILanguage';
```
pulkitvyas08, line 6: This should be a global import using path aliases

pulkitvyas08 (changes requested):

pulkitvyas08:
> This PR corresponds to #237. There are two problems I m facing: -
> 
> 1. There is a problem with the imports. Although I did try to add tsconfigpaths in tsconfig.json still it's showing a problem. Below is the snapshot. This problem is common for all the imports
>    image

This is not a real problem, these are just warnings, even I'm trying to get these fixed but to no avail as of now. I will create a dedicated issue for this later

> 2. And another one is this. Like how do I define the type for this RecoilState
>    image

For every recoil state, you need to define proper types. Go through the files in the recoil state and let me know if you need help with any particular one. Most of the states are pretty straightforward but in a few cases, you would have to change the structure of the states. Like in the case of `ThemeSelectionState` you would have to either go ahead and set the state to either string only or boolean only or write an interface to allow for both

pulkitvyas08:
Also I see a lot of conflicts that cannot be resolved from github's interface, have fun 😁

src/Interfaces/ILanguage.ts
Tushar-4781: will do. FYI I said *Kindly ignore the convention problems. It's not detected by the eslint. I have to check that part. Help me with the above 2 problems*

src/components/languagechoices/LangItem.tsx
Tushar-4781: path aliases ain't working that's why i have not used in the interfaces as well

Tushar-4781:
> Also I see a lot of conflicts that cannot be resolved from github's interface, have fun 😁

Umm, I can deal with the conflicts that are not the main issue. I will fix the conflicts by end of the weekend but the condition is that no one should be working on any new feature. What I mean is that once the code is 100% converted into TypeScript it can be merged then others can start working on that instead of working in JavaScript. What do you say @tijlleenders? Is it ok?

Tushar-4781:
@pulkitvyas08, You need to convert the MyFeelings feature into TS ( including the API calls and dexi database part)

I almost resolved all path aliases problems, but image imports are still showing warnings.

pulkitvyas08:
> @pulkitvyas08, You need to convert the MyFeelings feature into TS ( including the API calls and dexi database part)

Yes let's create a separate issue to deal with that plus TS compilation (that you haven't done, which is why the website is working without resolving all the TS related errors)

> I almost resolved all path aliases problems, but image imports are still showing warnings.

That is great to know, solves a lot of my problems

pulkitvyas08 (approved):

pulkitvyas08:
> > Also I see a lot of conflicts that cannot be resolved from github's interface, have fun 😁
> 
> Umm, I can deal with the conflicts that are not the main issue. I will fix the conflicts by end of the weekend but the condition is that no one should be working on any new feature. What I mean is that once the code is 100% converted into TypeScript it can be merged then others can start working on that instead of working in JavaScript. What do you say @tijlleenders? Is it ok?

In that case please go ahead and resolve these merge conflicts to get this PR merged. I hope the process doesn't create duplicate history though

src/components/myfeelingspage/AddFeelings.tsx
```diff
@@ -4,8 +4,13 @@ import { useTranslation } from 'react-i18next';
 import { useRecoilValue } from 'recoil';
 
 import { HeaderDashboard } from '@components/dashboard/HeaderDashboard';
+<<<<<<< HEAD:src/components/myfeelingspage/AddFeelings.tsx
```
pulkitvyas08: Why are these conflict resolution marks here?

pulkitvyas08 (changes requested):

src/components/myfeelingspage/AddFeelings.tsx
Tushar-4781: I m surprised. I resolved them at my local machine and then pushed. The working tree is clear, no conflicts are At my end

pulkitvyas08 (approved):

Tushar-4781:
Resolves #237

PR #290, opened 2022-05-07
Fixed the errors in TodoListForm...

GithinjiHans:


src/components/todoList/TodoListForm.tsx
```diff
@@ -19,25 +19,25 @@ export function TodoForm() {
...
+      id:Date.now().toString(),
       [e.target.name]: value,
     });
   };
 
-  const handleSubmit = (e) => {
+  const handleSubmit = (e:any) => {
     if (formInputData.inputGoal) {
-      const newData = (data) => ([...data, formInputData]);
+      const newData = (data:any) => ([...data, formInputData]);
```
Tushar-4781, line 34: we can't define type as any. It's an array and I think it's possible to give its type. @pulkitvyas08 what kind of array does db return?

Tushar-4781 (changes requested):

src/components/todoList/TodoListForm.tsx
pulkitvyas08: > we can't define type as any. It's an array and I think it's possible to give its type. @pulkitvyas08 what kind of array does db return or accept?

There is no db for goals yet. When the db is created we would need to make an API accordingly and there we can define what it would return or accept

sokorototo: Then I think `any` is actually a good fit here since the type is bound to change. Drop a _TODO:_ and change it in the future

sokorototo: Plus using `any` as a type for a callback is ok since you can't invoke it anywhere else. It's a once used, thrown away construct

sokorototo (approved):
`any` in a callback's parameter type is fine to have

pulkitvyas08:
@GithinjiHans ping me whenever you change the status of this PR to "ready for review"

GithinjiHans:
@pulkitvyas08  I've changed the PR's status to "ready to preview"...

src/components/todoList/TodoListForm.tsx
```diff
@@ -19,25 +19,25 @@ export function TodoForm() {
...
 
-  const handleChange = (e) => {
+  const handleChange = (e:any) => {
     const { value } = e.target;
-    const id = Date.now();
 
     setFormInputData({
       ...formInputData,
-      id,
+      id:Date.now().toString(),
```
pulkitvyas08, line 27: Why is the id value the current date? Why don't you use a counter or something instead? Maybe create a DexieJS collection and use that instead of manually defining everything. And please add some space between the colon and the text

pulkitvyas08 (changes requested):

src/components/todoList/TodoListForm.tsx
GithinjiHans: The id is calculated in milliseconds and they will never be the same when the user adds goals...I think its the most efficient id generator.

tijlleenders: Since we want to be able to distinguish events across many different users so that they can coordinate automatically I would suggest a UUIDv4 to make sure there is no conflict. A ms conflict is rare - but if you have millions of users - it will happen.

Agreed, the ms is very efficient - and fine for a single user - but in this case avoiding conflicts across users is more important than efficiency.

GithinjiHans: I've changed it now...I think the uuid was the best after all.

GithinjiHans:
This PR is ready @pulkitvyas08

pulkitvyas08 (approved):

pulkitvyas08:
@Tushar-4781 are there any changes that you are requesting? Your request is blocking the merge

Tushar-4781 (approved):

PR #307, opened 2022-05-10
Made the German word available in the listed languages

GithinjiHans:


Tushar-4781 (approved):

Tushar-4781 (commented):
Hey just a quick question, Aren't we supposed to create a subfolder for german in the translations directory?

PR #309, opened 2022-05-11
API implementation for Goals CRUD Operations  

### LINKED ISSUE #294: Make API to store goals
Make an API to store/query/delete goals - similar to the Feelings implementation.

Goal object should have an id, a title, and optionally a collection of addtitional key-value pairs (duration: x, repeat: string, start: datetime, finish: datetime, at: time, etc)

Tushar-4781:
Resolves #294

### DIFF
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -1,0 +1,44 @@
+import { db } from '@models';
+
+export const resetDatabase = () => db.transaction('rw', db.goalItems, async () => {
+  await Promise.all(db.tables.map((table) => table.clear()));
+});
+
+export const addGoal = (goalDetails :{
+  title: string,
+  duration: string,
+  repeat: string,
+  start: Date,
+  finish: Date,
+  createdAt: Date}) => {
+  db.transaction('rw', db.goalsCollection, async () => {
+    await db
+      .goalsCollection
+      .add(goalDetails);
+  }).catch((e) => {
+    console.log(e.stack || e);
+  });
+};
+
+export const removeFeeling = (goalId) => {
+  db.transaction('rw', db.goalsCollection, async () => {
+    await db.goalsCollection.delete(goalId);
+  }).catch((e) => {
+    console.log(e.stack || e);
+  });
+};
+
+export const getAllGoals = async () => {
+  const allGoals = await db.goalsCollection.toArray();
+  console.log(allGoals);
+  return allGoals;
+};
+
+export const getGoalsOnDate = async (date) => {
+  db.transaction('rw', db.goalsCollection, async () => {
+    const goalsList = await db.goalsCollection.where('start').equals(date);
+    return goalsList;
+  }).catch((e) => {
+    console.log(e.stack || e);
+  });
+};
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -11,7 +11,7 @@ export const addGoal = (goalDetails :{
   repeat: Repeat | string,
   start: Date,
   finish: Date,
-  at: Date}) => {
+  createdAt: Date}) => {
   const currentDate = getJustDate(new Date());
   const goals : GoalItem = { ...goalDetails, createdAt: currentDate };
   db.transaction('rw', db.goalsCollection, async () => {
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -1,4 +1,4 @@
-import { db, GoalItem } from '@models';
+import { db, GoalItem, Repeat } from '@models';
 import { getJustDate } from '@src/utils';
 
 export const resetDatabase = () => db.transaction('rw', db.goalItems, async () => {
@@ -7,8 +7,8 @@ export const resetDatabase = () => db.transaction('rw', db.goalItems, async () =
 
 export const addGoal = (goalDetails :{
   title: string,
-  duration: string,
-  repeat: string,
+  duration: Number,
+  repeat: Repeat | string,
   start: Date,
   finish: Date,
   at: Date}) => {
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -1,4 +1,5 @@
-import { db } from '@models';
+import { db, GoalItem } from '@models';
+import { getJustDate } from '@src/utils';
 
 export const resetDatabase = () => db.transaction('rw', db.goalItems, async () => {
   await Promise.all(db.tables.map((table) => table.clear()));
@@ -10,17 +11,19 @@ export const addGoal = (goalDetails :{
   repeat: string,
   start: Date,
   finish: Date,
-  createdAt: Date}) => {
+  at: Date}) => {
+  const currentDate = getJustDate(new Date());
+  const goals : GoalItem = { ...goalDetails, createdAt: currentDate };
   db.transaction('rw', db.goalsCollection, async () => {
     await db
       .goalsCollection
-      .add(goalDetails);
+      .add(goals);
   }).catch((e) => {
     console.log(e.stack || e);
   });
 };
 
-export const removeFeeling = (goalId) => {
+export const removeGoal = (goalId) => {
   db.transaction('rw', db.goalsCollection, async () => {
     await db.goalsCollection.delete(goalId);
   }).catch((e) => {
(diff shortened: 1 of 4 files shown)


Tushar-4781:
This same branch can be used for integrating the API calls on frontend

pulkitvyas08:
> This same branch can be used for integrating the API calls in frontend

No, different branches for different purposes. And please use more descriptive titles for branch names. `feature/add-goals` doesn't give the idea that you are just adding the API in this branch's PR

src/models/db.ts
```diff
@@ -1,13 +1,18 @@
...
   feelingsCollection!: Table<FeelingItem, number>;
 
+  goalsCollection!: Table<GoalItem, number>;
+
   constructor() {
     super('FeelingsDB');
     this.version(1).stores({
       feelingsCollection: '++id, content, category, date',
+      goalsCollection: '++id, title, duration, repeat, start, finish, at',
+
```
pulkitvyas08, line 15: Please remove the extra blank line here

src/models/GoalItem.ts
```diff
@@ -0,0 +1,9 @@
+export interface GoalItem{
+    id?: number;
+    title: string;
+    duration: string;
+    repeat: string;
+    start: Date;
+    finish: Date;
+    createdAt: Date;
```
pulkitvyas08, line 8: Do we really need this? There must be a native way in DexieJS to get an item's `createdAt` date. Please go through the API reference

src/api/GoalsAPI/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  db.transaction('rw', db.goalsCollection, async () => {
+    await db
+      .goalsCollection
+      .add(goalDetails);
+  }).catch((e) => {
+    console.log(e.stack || e);
+  });
+};
+
+export const removeFeeling = (goalId) => {
```
pulkitvyas08, line 23: Change the function's name to `removeGoals`

src/api/GoalsAPI/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  db.transaction('rw', db.goalsCollection, async () => {
+    await db.goalsCollection.delete(goalId);
+  }).catch((e) => {
+    console.log(e.stack || e);
+  });
+};
+
+export const getAllGoals = async () => {
+  const allGoals = await db.goalsCollection.toArray();
+  console.log(allGoals);
```
pulkitvyas08, line 33: Remove this console log

pulkitvyas08 (changes requested):

Tushar-4781:
> Great!
> 
> Couple of minor things:
> 
> * duration can be an int
> * repeat can be an enum (for now values once and daily)
> * at was meant as a time value, not a date like createdAt. 'at' is for giving a time preference to a goal, ie "lunch daily 1h at 13h". But I think createdAt is a useful field to have too.

Agreed. That's why I kept duration as a string like "lunch daily 1h at 13h" -> here **13h** is the duration write so if we keep int then we won't be able to provide the unit like hr, min, day

src/models/GoalItem.ts
```diff
@@ -1,9 +1,15 @@
+enum Repeat{
+    'Once',
+    'Daily'
+}
+
 export interface GoalItem{
     id?: number;
     title: string;
     duration: string;
```
tijlleenders, line 9: For now we assume all durations are in unit hours for the whole app.
If we do start using a unit specifier later - this will be in a separate field - and probably at calendar level so it doesn't have to be repeated in each goal.

For this reason, if you agree, please change duration to int without specifying the unit.

tijlleenders (commented):
Please use separate commits to resolve the different changes requested + good commit names.  

This makes it a lot easier to review. For example, if a change to function name and a change to a variable type were requested, a reviewer would at a glance see the commits:
- Fix function name
- Fix variable type
... and approve/refuse them independently of each other.

If we don't work this way, then some changes might be acceptable, while others aren't, and the result is a monolithic 'changes requested' => have to go through all the comments to see which ones are not resolved or mention it in the review comments => submit review again => reviewer goes through all the review comments again to check which one was not accepted... you get the picture.

src/models/GoalItem.ts
Tushar-4781: okay.

Tushar-4781:
Quick Question: why only 2 checks are there in 'ALL' ?

pulkitvyas08:
> Quick Question: why only 2 checks are there in 'All checks have passed'?

This is indeed strange, I'll try to get to the bottom of it but for the time being just let it be since the tests will run on merge too

src/api/GoalsAPI/index.ts
```diff
@@ -0,0 +1,46 @@
+import { db, GoalItem } from '@models';
+import { getJustDate } from '@src/utils';
+
+export const resetDatabase = () => db.transaction('rw', db.goalItems, async () => {
+  await Promise.all(db.tables.map((table) => table.clear()));
+});
+
+export const addGoal = (goalDetails :{
+  title: string,
+  duration: string,
```
pulkitvyas08, line 10: Are you sure we want the duration to be a `string`?

src/api/GoalsAPI/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  repeat: string,
```
pulkitvyas08, line 11: A goal is either repeating or not repeating. Why don't we use a boolean here?

src/api/GoalsAPI/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  start: Date,
+  finish: Date,
+  at: Date}) => {
```
pulkitvyas08, line 14: What is this `at`? Were you able to find a createdAt field for database items in DexieJS?

src/api/GoalsAPI/index.ts
pulkitvyas08: Oh I see that there is already an enum to be used for repeat. Please use that instead of a boolean

src/models/GoalItem.ts
pulkitvyas08: Why did you resolve this conversation? Did you find a way to do this?

src/models/db.ts
```diff
@@ -1,13 +1,17 @@
 import Dexie, { Table } from 'dexie';
 import { FeelingItem } from './FeelingItem';
+import { GoalItem } from './GoalItem';
 
 export class FeelingsDB extends Dexie {
   feelingsCollection!: Table<FeelingItem, number>;
```
pulkitvyas08, line 7: Remove this blank line here

pulkitvyas08 (changes requested):

Tushar-4781:
My apologies for repeating mistakes, I will take care of these changes.

src/api/GoalsAPI/index.ts
Tushar-4781: I missed it

src/api/GoalsAPI/index.ts
Tushar-4781: No, I will create an issue on this.
I made an alternate for now. **createdAt** will act as a timestamp. 
**at** is removed

pulkitvyas08: @Tushar-4781 there is still an `at` here at line 14. Please change that to `createdAt too`

Tushar-4781: I made an alternate for now. **createdAt** will act as a timestamp.

src/models/db.ts
Tushar-4781: Actually, I can't because eslint is throwing an error **Expected blank line between class members**

src/models/GoalItem.ts
pulkitvyas08: Since we haven't been able to find a way to change this to a native DexieJS value, let's go ahead and get this merged and then we can take this up shortly. @Tushar-4781 create an issue regarding the same

pulkitvyas08 (changes requested):

src/api/GoalsAPI/index.ts
Tushar-4781: > @Tushar-4781 there is still an `at` here at line 14. Please change that to `createdAt too`

Kindly check the second commit. It is dedicated to that

pulkitvyas08 (approved):

tijlleenders (approved):

src/models/GoalItem.ts
tijlleenders: Created #317 for this

PR #311, opened 2022-05-12
Adding Brazilian/Portuguese language

ubermenzchen:
Brazilian/Portuguese translations.
Note: the (a)'s at the end of a noun denotes genre. Ending with 'o' is masculine and with 'a' is feminine.


src/components/landingpage/UserChoicePanelRight.tsx
```diff
@@ -41,6 +41,11 @@ export function UserChoicePanelRight() {
       title: t('german'),
       langId: 'de',
     },
+    {
+      sno: 7,
+      title: t('portuguese/Brazil'),
```
tijlleenders, line 46: Punctuation should follow the same scheme as the other languages.
So "Portuguese" would be better.

tijlleenders (changes requested):
Thanks Matheus!
...and welcome to ZinZen.

A small fix and then we can merge.
I guess for future contributors we'll have to be more specific in the issue description.
Adding a translation file is not the same thing as wiring it up in the frontend.
The https://zinzen-git-fork-ubermenzchen-add-a-new-lang-60ea45-tijlleenders.vercel.app/ shows your code - and the Portuguese option is not working yet.

tijlleenders:
I converted back to draft.
When you're done, upgrade it back to a PR status so we can clearly see when a reviewer's attention is required again.

Thanks!

src/components/landingpage/UserChoicePanelRight.tsx
tijlleenders: I have no idea - but would a Brazilian expect to see "Brazilian" or "Portugese"?

Please make a choice and capitalize the first letter so it's consistent in the UI.

tijlleenders:
Nice!
The translations now work in the frontend.

tijlleenders:
#379 takes these changes 
+ resolves conflicts
+ adds translation for "Portuguese" as a language in the translation files.

PR #314, opened 2022-05-13
added Pull Request from fork actions triggers

tijlleenders:
As can be seen in PR #311 the GitHub Actions don't get triggered.
This is because the PR comes from a forked repo  
  => not pushing to the ZinZen repo  
  => not triggering 'push' event on ZinZen repo
  => not triggering GitHub Actions.

I added a trigger for pull_request: [opened, edited, ready_for_review]

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -1,5 +1,8 @@
 name: Cypress
-on: [push]
+on:
+  push:
+  pull_request: [opened, edited, ready_for_review]
+
 jobs:
   cypress-run:
     runs-on: ubuntu-latest
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   cypress-run:
     runs-on: ubuntu-latest
+    if: github.event_name != "pull_request" || github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


tijlleenders:
Hmm even for this PR, only 1 check is running...
@pulkitvyas08 thoughts?

tijlleenders:
I see some errors in the Actions. Let met fix...

tijlleenders:
Need a fix so that Actions only run once if triggered by multiple events:
image

Adding this line as a condition to the job does just that:
```
if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
```
The job will still try to run twice - but the if will abort one of them before running.  
Found the solution on [this page](https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/)

We'll have to merge the code first to double-check it also has the required behavior for PRs from forked repos.

tijlleenders:
Hmmm... it's still running twice.
Will have to do some more research.

tijlleenders:
So the first time I changed the workflow the actions got run twice in the PR overview...

But now it's OK:
image


Confirmed they run only once in the Actions overview:
image

tijlleenders:
Ah...
I understand now - it's not that they're running twice on a single commit.  
It's just that they run:
- once when I commit to the PR
- then also once when I set the PR from draft to final status.
The PR UI then shows 7 checks.

So the confusion is cleared up!

pulkitvyas08 (approved):

PR #325, opened 2022-05-14
Added a parser for daily and modified the appearance of the change co…

GithinjiHans:
…lor button


tijlleenders (changes requested):
The test doesn't yet take/return a goal object as described in the issue.

tijlleenders (approved):
Think @Tushar-4781 should review too before we merge

Tushar-4781 (approved):

Issue #312, opened 2022-05-12
Show saved feelings on My feelings page


tijlleenders:
Blocked by #313 

- Make a My feelings page as designed in #313
- The feelings get saved when you click on them on the Add feelings page.
- After the Feeling is saved, the user needs to be taken to the 'My feelings' page (where the Feeling is shown) so that it is clear that the Feeling has been added.

pulkitvyas08:
> * After the Feeling is saved, the user needs to be taken to the 'My feelings' page (where the Feeling is shown) so that it is clear that the Feeling has been added.

I think a better way to achieve this would be by showing a toast message to the user that the feeling has been added or that the feeling already exists in today's logs. Attaching a sample image to show what I mean by this.

![image](https://user-images.githubusercontent.com/54947769/168674229-8238c183-fadb-4dfe-a1a0-00c218867abf.png)

tijlleenders:
Blocked by #313 that still needs to be developed.

tijlleenders:
Good suggestion.

Let's keep the current flow so we can ship a first "working" version. 
If we confirm that people find it confusing that they are taken to the My feelings page as confirmation after adding a feeling we can explore this further.

tijlleenders:
Design in #313 is finalized.

PR #339, opened 2022-05-16
Feature/render feelings

pulkitvyas08:
Part of #312 

I mentioned in the last meeting that displaying stored feeling is working but then checking with the deployed dev I found that I never pushed the fix so here it is. 

Also fixed the emoji issue on add feelings page (that no one caught till now AFAIK, showing that the test coverage of the UI is pretty bad at the moment)

### DIFF
--- a/src/api/FeelingsAPI/index.ts
+++ b/src/api/FeelingsAPI/index.ts
@@ -1,20 +1,9 @@
 import { db } from '@models';
 import { getJustDate } from '@src/utils';
 
-export const resetDatabase = () => db.transaction('rw', db.feelingItems, async () => {
+export const resetDatabase = () => db.transaction('rw', db.feelingsCollection, async () => {
   await Promise.all(db.tables.map((table) => table.clear()));
 });
-
-export const addFeeling = (feelingName : string, feelingCategory : string) => {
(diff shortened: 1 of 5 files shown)


tijlleenders (changes requested):
Good catch! Agree that it is time to start using auto UI tests.
We can announce and explain this in the next meeting.

We previously set up Cypress but we're not using it. I made #341 to create a list of the UI tests that we'll need and how to integrate them to the development process. It should be possible to run them locally but not auto-run on every local commit.

Tested this manually - and for mobile (we design develop for mobile only at the moment) the emoji's are not rendering.

If you fix that I can merge.

tijlleenders:
This PR is useful - but will not close #312

pulkitvyas08:
> This PR is useful - but will not close #312

Yes, that was never the intention. I wrote 

> Part of https://github.com/tijlleenders/ZinZen/issues/312

Above instead of `resolves #312`

pulkitvyas08:
> Tested this manually - and for mobile (we design develop for mobile only at the moment) the emoji's are not rendering.
> 
> If you fix that I can merge.

Yupp, my bad, never checked it for mobile

That is fixed now though

tijlleenders (approved):

Issue #273, opened 2022-05-06
Show suggestions as goals are typed


tijlleenders:
#270 and #272 should be closed before starting this.

Show suggestions as small badges whenever the goal parser function from #270 has a suggestion => show it like designed in #213

PR #344, opened 2022-05-18
Made the daily suggestion available when user types in daily

GithinjiHans:


tijlleenders (approved):
The badge appears when typing.
The appearance of the badge is not yet optimal, but that is ok as it will be fixed in #273

PR #359, opened 2022-05-21
Fix errors in UserChoice.test file

### LINKED ISSUE #349: Errors in UserChoice.test.ts
1. There are eslint errors

Screenshot (16)

&nbsp;

2. There are TS errors (#347)

Screenshot (16)

&nbsp;

3. Running the test doesn't even do anything

&nbsp;

Screenshot (18)

&nbsp;

4. Not sure about what the test is about. Please give a description.

Assigning the people who worked on this.

Tushar-4781:
Resolves #349 

For now, I have fixed the pre-designed test cases in UserChoice.test file. They are running now. However, the Path aliases problem has to be fixed for the Unit and Integration test files as well and for that we need to make a configuration file for jest. This thing can be done in the issue where we are developing the approach and list of UI tests.
( PS: Can anyone tag that issue )


src/__tests__/UserChoice.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+});
+
+describe('Language Chosen', () => {
+  test('check chosen theme', () => {
+    const chosenLang = localStorage.getItem('language');
+    if (chosenLang) { expect(languagesAvailable).toContain(chosenLang); }
+  });
+});
+// const theme = localStorage.getItem('theme');
```
pulkitvyas08, line 17: Delete these comments if they are not needed

pulkitvyas08 (approved):
I see that the `App.test.ts` file gives errors but do we really need to remove it altogether or is there a way to fix those problems. Anyway since this is not in the scope of this PR, I'm approving it

src/__tests__/UserChoice.test.ts
Tushar-4781: Almost forgot 😅

pulkitvyas08 (approved):

PR #363, opened 2022-05-22
Wire up the duration parser to front end

GithinjiHans:


tijlleenders (approved):
Good first version. This is the idea.

I'll have some extra (edge) test cases soon - and there is some extra formatting to do.
Those are for later as separate issues.

Not sure about code style. That should be automated, so we don't have to think about it.

tijlleenders (approved):

PR #367, opened 2022-05-24
Made the Header static

GithinjiHans:


tijlleenders (changes requested):
Header still scrolls with page.

GithinjiHans:
In which page?...I might have left it out

tijlleenders:
The first page I looked at (startpage):
image

GithinjiHans:
The start page has a different header.In the othe pages the Header doesn't scroll with the page.

PR #377, opened 2022-05-28
MyGoals Page

### LINKED ISSUE #352: Make 'My goals' page
Show added goals like in design #302

Tushar-4781:
Resolves #352


tijlleenders (changes requested):
Can't add any goals so can't see My Goals either.

Tushar-4781:
- We Need to discuss and finalize how the Add-in Sublist, Edit, and Checked buttons will work.
- For now, certain assumptions are made:
    - If we click on the title it becomes editable as soon as focus is lost, the new title will be saved in DB
    - Delete button will simply delete the goal

Bug: If the focus is on the title then we can click anywhere outside the goal div and the options will become invisible. But apart from that like if we click somewhere in div apart from the title and then click outside then options are not becoming invisible. I tried my best but for this case, the Lost Focus property isn't working. Give me a hint if you have a solution to this or I will create a quick small issue for this. 

Rest I have covered almost every edge case. Try it out and lemme know if any fix is reqd.

Tushar-4781:
> Can't add any goals so can't see My Goals either.

I have created a function for adding a dummy data. Do you want me to uncomment and push it or if you want you can do it at your end as well.  just call that function in useEffect

tijlleenders (changes requested):
I uncommented the dummy code and added it to useEffect to be able to test.

It is coming together :)

1. It may be a small thing ... but can you use the theme colors?
2. Can you stop the colors from changing on every click? It is quite confusing.
3. The first click should just expand to show the buttons - only the second single click should edit.
4. I feel the height of the goals is a little smaller than in the design. Can you increase a bit? Also for the room around the icons. Currently pretty tight to the title text and the border.
5. Missing 'My goals' at the top as in design
6. The icons are different than the design. Bootstrap-icons are nice, but they are a little off theme in terms of thickness. @mallikarai05 Can you say which ones you used for the design?

tijlleenders:
7. The area used for the display can be increased. Some screen real estate is left unused:

image

Tushar-4781:
> 

1. & 2. --> Actually, I didn't have the list of colors that has to be used. So the random color generator is generating the colors and since the states are updating the re-rendering is causing the change in colors. 

3.  It is designed that way only it just the touch on title and div containing are overlapping. I have found a new way cuz it seems like the focus attributes of an HTML element can't help us directly. Certainly, we need to design a specific handler for this 2-tap process. Will Create and push in some time.

4. & 5. Sure, I will increase the height and will make my goals title visible. Sorry, My mistake I wasn't working on the latest code. I forgot the header was modified. 

6. Yeah, although I can download those icons as images but IDK which way is more optimal downloading the icon images or importing them from some library. @mallikarai05 please share the icons library.

7. I kept like that because I had a doubt about how users will know if there are more goals on the list? actually there was space given in the design between the list and the add button.

mallikarai05:
@Tushar-4781 I actually the icons in the design are from icons8, I can make a zip file and send them to you?

Tushar-4781:
> @Tushar-4781 I actually the icons in the design are from icons8, I can make a zip file and send them to you?

sure.

tijlleenders:
1 & 2 : you can look for the current colors in the code. Just hardcode a single color for now as we only have one list (no sublists) and now way to change the color yet.

tijlleenders:
3 Don't spend too much time on it. The touch event probably get propagated. Stopping the propagation im combination with the right order (z-index) can lead to complicated logic quite fast.

tijlleenders:
6 Will the unused icons get excluded from the final build?

tijlleenders:
7 Good point Tushar. I'll make an issue to include a spacer at the bottom of every page so when you scroll down it is obvious that it is the end of the list. For now just let the view of the list go all the way to thr bottom of the page.

tijlleenders:
Fantastic!

I'll make a separate issue for the icons.

tijlleenders (approved):

Issue #321, opened 2022-05-14
Make a 'Download my data' button/option


tijlleenders:
No UI necessary.
For now just a function that can be called from the javascript web developer console to backup the whole repo to a file on the user's device.

The file is not encrypted at the moment - there will be a separate issue for that.

GithinjiHans:
We can do this using a ShellScript

tijlleenders:
What is wrong with this?
![image](https://user-images.githubusercontent.com/44887755/171099893-14a43929-a6f1-4617-86f8-ee311dea1503.png)

I don't see the added value of a shell script.


What happened here is that you misinterpreted the issue.
It is about making a file download of all registered feelings and goals.

tijlleenders:
This issue is not a priority for now.

tijlleenders:
It also needs a design (download button in settings menu) and a file format spec.

PR #382, opened 2022-05-30
Made the download repo shell script.

GithinjiHans:
To run it on terminal type ./downloadRepo.sh

### DIFF
--- a/downloadRepo.sh
+++ b/downloadRepo.sh
@@ -1,0 +1,1 @@
+git clone https://github.com/tijlleenders/ZinZen.git
--- a/downloadRepo.sh
+++ b/downloadRepo.sh
@@ -1,1 +1,3 @@
+cd ~/Downloads
+cd \/Downloads
 git clone https://github.com/tijlleenders/ZinZen.git


GithinjiHans:
I haven't tested it on windows...anyone using windows test it and give me your feedback

pulkitvyas08 (changes requested):
Doesn't work on windows. There's no `~/` in Windows

tijlleenders:
I don't understand what issue this PR closes.
If it's just a one-line shell script I don't think it adds anything.

GithinjiHans:
It closes issue #321 but we can use a button and this link https://github.com/tijlleenders/ZinZen.git in the UI for the download repo function but issue #321 says that it should run in console so a shell script is the best solution.

tijlleenders:
Deleting this PR as it was based on a misinterpretation of the issue #321

tijlleenders:
> It closes issue #321 but we can use a button and this link https://github.com/tijlleenders/ZinZen.git in the UI for the download repo function but issue #321 says that it should run in console so a shell script is the best solution.

Sorry, I meant the javascript web developer console.

Issue #328, opened 2022-05-15
Avoid empty My feelings page


tijlleenders:
If someone opens the My feelings page without any registered feelings it should open the Add feelings page instead.

Issue #318, opened 2022-05-13
Make optional note popup when logging feeling


tijlleenders:
Make the popup to ask for a reason when adding feeling like in #276 .
If dismissed (ie by clicking outside the popup area) the feeling still gets stored.

PR #384, opened 2022-05-31
Feature/my feelings actions

### LINKED ISSUE #282: Click on tile to delete feeling
Can be achieved using the `removeFeeling` function in the `feelingsAPI`

### LINKED ISSUE #312: Show saved feelings on My feelings page

pulkitvyas08:
**WIP!**

Resolves #282
Resolves #312 

@Tushar-4781 please take a look at this. `ShowFeelings.tsx` and `ShowFeelingsTemplate.tsx` contain the code that is getting into the way of async rendering of the deleted items. If you want then you can add the commit for performant rendering to this PR itself


src/pages/ShowFeelingsPage/ShowFeelingsPage.tsx
```diff
@@ -48,8 +56,9 @@ export const ShowFeelingsPage = () => {
               </h3>
               <ShowFeelingTemplate
                 key={feelingsList[date]}
-                feelingCategory="Test"
                 feelingsListObject={feelingsList[date]}
+                setFeelingsListObject={handleFeelingsListChange}
```
Tushar-4781, line 60: Do not handle/pass the parent state like this if you want to change it from child.
It should be                 
setFeelingsListObject={{ feelingsList, setFeelingsList }}

Tushar-4781:
@pulkitvyas08 The above changes will surely solve the problem.

src/pages/ShowFeelingsPage/ShowFeelingTemplate.tsx
```diff
@@ -1,59 +1,70 @@
...
-                key={feelingsListObject[feelingObject].content}
+                key={feelingsListObject[Number(feelingId)].content}
                 className={
                 darkModeStatus ? 'btn-my-feelings-dark btn-feelings-dark' : 'btn-my-feelings-light btn-feelings-light'
               }
                 size="lg"
-                onClick={() => console.log('TODO: Add remove feeling function')}
+                onClick={() => {
+                  removeFeeling(feelingsListObject[Number(feelingId)].id);
+                  let newFeelingsList = currentFeelingsList;
```
Tushar-4781, line 39: Suggestion : In Showfeelings page, you already calling getfeelings in useeffect. Make a function of that in the same file. 
Use that here. This would reduce the lines of code because anyways you are close to O(n) complexity in both cases i guess.

Tushar-4781 (changes requested):

src/pages/ShowFeelingsPage/ShowFeelingsPage.tsx
pulkitvyas08: Thanks for the suggestion! Actually, this is the way to do it when you are using something like recoil for state management of these lists i.e., this is the performant way mentioned in react docs. But for props, it doesn't really work apparently.

src/pages/ShowFeelingsPage/ShowFeelingTemplate.tsx
pulkitvyas08: This is for modularization. One is the template and another is the populator so I don't think it matters anyway

src/pages/ShowFeelingsPage/ShowFeelingsPage.tsx
Tushar-4781: well, let's see if we encounter any such scenario in the future. here it works at least. ;)

pulkitvyas08:
Resolves #318
Resolves #328

pulkitvyas08:
Just observed that there is an issue with My Feelings page incorrectly redirecting all users to Add Feelings page. Blocking merge till I fix that

pulkitvyas08:
PR is ready for review and merge! I have taken care to make sure that it doesn't break the older version of the database (that doesn't have a `notes` field) but let me know if that becomes an issue

tijlleenders:
Yeah!
This is a major improvement in UX - and adds a great feature (note taking).

tijlleenders (changes requested):
Some minor changes to finish up.
1. The popup is not themed to dark in dark mode - it should be
2. The popup does not focus on text input right away - it should
3. Pressing enter in the popup does not submit the note - it should
4. When adding a note, the component does not show the note indication ('...') until you navigate away and come back - it should refresh the component as soon as the state changes

pulkitvyas08:
> Some minor changes to finish up.
> 
>     1. The popup is not themed to dark in dark mode - it should be

Design for this wasn't in the mentioned issue so I forgot to add, will do that.

>     2. The popup does not focus on text input right away - it should

Good catch, will make sure that happens

>     3. Pressing enter in the popup does not submit the note - it should

My bad, I thought I did add a keydown but apparently that keyDown was for another component

>     4. When adding a note, the component does not show the note indication ('...') until you navigate away and come back - it should refresh the component as soon as the state changes

@Tushar-4781 facing the same issue for async rendering here. Can you chime in and see what can be done to fix this. Refer to `ShowFeelingTemplate` in `src/pages`, thanks

src/pages/ShowFeelingsPage/ShowFeelingTemplate.tsx
```diff
@@ -1,59 +1,224 @@
...
+                  className="show-feelings__note-input"
+                  value={noteValue}
+                  onChange={(e) => { setNoteValue(e.target.value); }}
+                />
+              </Modal.Body>
+              <Modal.Footer>
+                <Button
+                  variant="primary"
+                  onClick={() => {
+                    addFeelingNote(selectedFeeling!, noteValue);
```
Tushar-4781, line 186: Well, When you are adding note, you are directly updating the Dexi DB. My suggestion is after updating the DB return the updated feelings from the DB. So update the API call after adding the note in DB return the updated data from there and change the state.

src/api/FeelingsAPI/index.ts
```diff
@@ -29,22 +37,43 @@ export const getFeelingsOnDate = async (date: Date) => {
...
+};
+
+export const addFeelingNote = async(feelingId: number, InputNote: string) => {
+  const feeling = await db.feelingsCollection.get(feelingId);
+  const updatedFeeling = {...feeling, note: InputNote};
+  db.transaction('rw', db.feelingsCollection, async () => {
+    await db.feelingsCollection.put(updatedFeeling);
+  }).catch((e) => {
+    console.log(e.stack || e);
+  });
```
Tushar-4781, line 69: Right after this line make a call get the updated feelings and return

Tushar-4781 (changes requested):

src/pages/ShowFeelingsPage/ShowFeelingTemplate.tsx
pulkitvyas08: Yes that's what I tried before but did not get the re-rendering successful. Let me go back to that commit and post it here for your reference

src/api/FeelingsAPI/index.ts
pulkitvyas08: Exactly what I did

tijlleenders (dismissed):
Yup, all good!

src/pages/ShowFeelingsPage/ShowFeelingTemplate.tsx
```diff
@@ -1,59 +1,285 @@
...
+  const [noteValue, setNoteValue] = useState('');
+
+  const handleInputClose = () => setShowInputModal(false);
+  const handleInputShow = () => setShowInputModal(true);
+  const handleNotesClose = () => setShowNotesModal(false);
+  const handleNotesShow = () => setShowNotesModal(true);
+  const handleFeelingsNoteModify = async () => {
+    addFeelingNote(selectedFeeling!, noteValue)
+      .then((newFeelingsList) => {
+        // @ts-ignore
```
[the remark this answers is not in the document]
pulkitvyas08: https://github.com/Microsoft/TypeScript/issues/24587 Let me know if you find a way to resolve this error `Type 'Date' cannot be used as an index type`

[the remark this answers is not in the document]
pulkitvyas08: Not working, creating additional errors too. The logic seems fine and since there is no extension at the moment the priority can be shifted to release rather than fix these problems. I will create an issue for this but let's get this into production for now.

Tushar-4781 (approved):

PR #387, opened 2022-06-01
Enabling Install to Desktop

Tushar-4781:
WIP!!

Resolves #207

PR #388, opened 2022-06-03
231 service worker

GithinjiHans:


Tushar-4781:
I can feel the pain @GithinjiHans ;D
This will be the first PR of my life where I see 45 commits. Cheers... lol ....

src-sw.js
```diff
@@ -2,4 +2,8 @@ importScripts('https://storage.googleapis.com/workbox-cdn/releases/3.0.0/workbox
 
 console.log('service worker customized');
 
+workbox.routing.registerRoute(
+    new RegExp('https://zinzen-qqu97eop7-tijlleenders.vercel.app/Home/MyGoals),
```
Tushar-4781, line 6: I think you are missing the quote ( ' ). the URL is not in quotes

tijlleenders:
Maybe use a [vite plugin](https://vite-plugin-pwa.netlify.app/) to make this work if the [public directory](https://vitejs.dev/guide/assets.html#the-public-directory) doesn't fix it?

There's also a [working example PWA](https://github.com/imranhsayed/react-pwa-workbox-example) to compare with.

tijlleenders:
Can #387 be closed and deleted in favor of this one?

PR #392, opened 2022-06-04
Dashboard Tests

### LINKED ISSUE #390: Write Tests for DashboardUserChoicePanel
DashboardUserChoicePanel is fully ready according to design. It's a good start for writing the test cases. Both Cypress and Jest test cases can be written for this page.

Tushar-4781:
WIP!!

Resolves #390

Cypress Tests are written and Just a quick question can we downgrade to React 16 because for jest I want to use enzyme and its **official version** is not released for react 17. Although there is one unofficial version for react 17 so what do we do? downgrade react or use that enzyme unofficial version?
@tijlleenders @GithinjiHans @pulkitvyas08 What are your thoughts?


tijlleenders:
Whybfo you need enzyme?

Tushar-4781:
> Why do you need enzyme? Are there no alternatives that do the same for React 18?

It comes handy with Jest and a nice combination to use. Enzyme provides a mechanism for mounting and traversing React.js component trees.  You can easily assert, manipulate, and traverse components using the framework. Unlike in the past, when you might not have had access to the renderer, Jest can render the content, whereas Enzyme allows you to test assumptions about the content.

Tushar-4781:
> Why do you need enzyme? Are there no alternatives that do the same for React 18?

For now i just want to use its mount function.

tijlleenders:
[Second Google result for 'enzyme react 18'](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl)

Maybe we should find a different way for testing... 
There must be something suitable for 18 right?

Tushar-4781:
There are no possible Jest test cases for this page.

pulkitvyas08 (approved):

PR #397, opened 2022-06-06
Service worker/vite plugin

GithinjiHans:


GithinjiHans:
The service worker is currently working but it's not caching the images....I'll try working something out

tijlleenders (changes requested):
Great to hear the service worker is finally working!

A little too enthusiastic to push the /dist folder too right? :)

GithinjiHans:
the dist folder is required because now it's linked with the index.html file in root.

GithinjiHans:
Now you can try reloading any page while offline....Now ZinZen can be used offline without breaking

GithinjiHans:
@tijlleenders the change requested can't be sollved for now because /dist is linked with the root html

pulkitvyas08:
@Tushar-4781 @GithinjiHans What is the end goal of this PR? Like how can I evaluate whether to approve it or not?

Tushar-4781:
Actually @GithinjiHans is trying to install services worker that can work on deployed url as well so if that's done. You can approve it. I gave up on this issue now.😅 I have tried everything I knew. @GithinjiHans still trying something.

pulkitvyas08:
> @pulkitvyas08 Actually @GithinjiHans is trying to install service worker that can work on deployed url as well so if that's done. You can approve it. I gave up on this issue now.😅 I have tried everything I knew. @GithinjiHans still trying something.

@GithinjiHans?

pulkitvyas08:
@Tushar-4781 sometimes there are also issues related to the production vs preview. Vercel treats them differently in a few aspects. I don't think that is the issue but a possibility to explore. Maybe try to do this on your fork by setting up vercel there

Tushar-4781:
> @Tushar-4781 sometimes there are also issues related to the production vs preview. Vercel treats them differently in a few aspects. I don't think that is the issue but a possibility to explore. Maybe try to do this on your fork by setting up vercel there

@pulkitvyas08 yeah, that's the reason i m not creating any new PR and closed the old ones

tijlleenders:
> the change requested can't be sollved for now because /dist is linked with the root html

Could you explain a little - I don't understand.

tijlleenders:
@GithinjiHans Can we exclude the workbox modules we don't need?
    "workbox-background-sync": "^5.1.4",
    "workbox-broadcast-update": "^5.1.4",
    "workbox-cacheable-response": "^5.1.4",
    "workbox-core": "^5.1.4",
    "workbox-expiration": "^5.1.4",
    "workbox-google-analytics": "^5.1.4",
    "workbox-navigation-preload": "^5.1.4",
    "workbox-precaching": "^5.1.4",
    "workbox-range-requests": "^5.1.4",
    "workbox-routing": "^5.1.4",
    "workbox-strategies": "^5.1.4",
    "workbox-streams": "^5.1.4",

GithinjiHans:
> @pulkitvyas08 Actually @GithinjiHans is trying to install service worker that can work on deployed url as well so if that's done. You can approve it. I gave up on this issue now.sweat_smile I have tried everything I knew. @GithinjiHans still trying something.

If we remove the /dist folder the app wont even be deployed registerSW.js is linked with index.html in the root directory....I couldn't find any other way of doing it

GithinjiHans:
> @Tushar-4781 @GithinjiHans What is the end goal of this PR? Like how can I evaluate whether to approve it or not?

The end goal is to make sure the service worker is working as required and thats done so I think we're good to go

tijlleenders:
@GithinjiHans are we using stale-while-revalidate strategy?

I don't want to merge the /dist.
There must be a better way.

GithinjiHans:
It's now using stale while revalidate

GithinjiHans:
> @GithinjiHans are we using stale-while-revalidate strategy?
> 
> I don't want to merge the /dist. There must be a better way.

Stale-while-revalidate isn't efficient enough....u'll notice that on offline mode it just crashes

tijlleenders:
We can't merge this.

"It works" is not enough. 
It needs to be clean too.

I'm closing the PR so we can focus on something else.

GithinjiHans:
Offline-first implementation was better

tijlleenders:
I understand your (and Tushar's) frustration. I also see no reason why we wouldn't be able to implement this cleanly when other projects can.

I prefer offline-first too!
... but we initially have to download the app and also keep it updated, right?

What strategy are you referring to as "better" than stale-while-revalidate in the workbox strategies - and why is that more suitable?

I'm open to pure offline mode. How do we keep the app updated if it is pure offline?

tijlleenders:
From the video you mentioned in direct message chat: https://youtu.be/cmGr0RszHc8?t=1774 
"With offline-first we get stuff from the cache first, and then we try and get content from the network".

That at least looks like its similar to stale-while-revalidate. 
...Maybe there's some difference in how it handles a failure from the network but I don't think so.

I'll have a go at it on Sunday. If it doesn't work out we'll find outside help.

PR #403, opened 2022-06-07
Fix: Design of Explore Page

### LINKED ISSUE #402: Fix Explore Page design
Explore page design is not matching with the Figma board. 
This is how it looks like rn
image

Tushar-4781:
Resolves #402


tijlleenders (approved):
Can't see if there are any warnings (on mobile) but looks good.

Issue #231, opened 2022-04-21
Make the App work offline with service worker


mallikarai05:

tijlleenders:
Read [Mozilla howto](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Installable_PWAs) and [Google howto](https://web.dev/customize-install/) for guidance.

tijlleenders:
Closed by #413

Issue #208, opened 2022-04-08
Enable install to Android


tijlleenders:
A PWA should be installable to Android.

tijlleenders:
Read [Mozilla howto](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Installable_PWAs) and [Google howto](https://web.dev/customize-install/) for guidance.

tijlleenders:
Closed by #413

Issue #207, opened 2022-04-08
Enable install to desktop


tijlleenders:
Using Chrome a PWA should be installable to desktop.

tijlleenders:
Pausing dev on this till I figure out how to continue with #231

tijlleenders:
PR #413 solves this in production (zinzen.me):
![image](https://user-images.githubusercontent.com/44887755/173285100-ab23f1c5-fec3-425c-82b7-f93a475fd29e.png)

...but the Vercel preview doesn't show the install promt in the browser.

tijlleenders:
Closed by #413

PR #413, opened 2022-06-12
Feature/stale while revalidate

tijlleenders:
@GithinjiHans @Tushar-4781 

Gave the offline-first (stale-while-revalidate) caching for offline mode a go.
TLDR: Local is working as well as the manually deployed zinzen.me :)

image

Will need some help to make this PR clean and configure (pre-)caching details and all - but I think the basics are here.

I used the create-react-app PWA template as a basis for the registerSW and service-worker files.
This is I believe also the basis @mallikarai05 used for initiating this codebase.

I first tried withoug the Vite-PWA plugin...
The build/bundle process makes things more complicated as now we have to instruct it to also copy the service worker file - but not add a hash to it? Or maybe we do want the hash for that too? Anyway - I couldn't configure it right even though it is quite [well documented](https://rollupjs.org/guide/en/). 
This is the downside of using a builder/bundler: it creates another tool to learn and you can't easily relate source files to deployed files.
I left the rollup code commented out. If it is glaringly obvious what I'm doing wrong please teach me :)
image

Then I tried using the Vite-PWA plugin which nearly worked immediately.

So we now need to choose to go the VitePWA route or stick with Rollup to avoid yet another dev dependency/plugin.


pulkitvyas08 (changes requested):
VitePWA seems to be a good choice. @Tushar-4781 can you look into fixing the cypress errors for this PR and maybe extend this work to add offline installation enabled for desktop, even if that doesn't get merged with this PR we will get an idea if this setup is useful or not

tijlleenders:
Good idea.

I just checked for mobile and there it seems to be installing. We might want to make a separate issue to catch the install event so we can couple it with a button in the menu.

pulkitvyas08:
> 

But you are testing on [zinzen.me](https://zinzen.me/). For this PR's preview URL I'm getting this error `Error during service worker registration: TypeError: ServiceWorker script at https://zinzen-git-feature-stale-while-revalidate-tijlleenders.vercel.app/service-worker.js for scope https://zinzen-git-feature-stale-while-revalidate-tijlleenders.vercel.app/ threw an exception during script evaluation.`

tijlleenders:
There is something I removed that might be able to solve this. 

The two base files you get when using create React app (CRA) with template PWA for Typescript mention a public URL env variable - but I removed that since it was undefined when running on localhost.

tijlleenders:
Oh wait - they're identical already.
Hmmm... that could mean trouble.

tijlleenders:
@pulkitvyas08 I get the same error when building on localhost:
```
Uncaught non-precached-url: non-precached-url :: [{"url":"undefined/index.html"}]
```

I had solved it by adding this to the Vite-PWA plugin - but forgot to commit the fix:
```
      workbox: {
        cleanupOutdatedCaches: true, //default is true but added to make it explicit
        ignoreURLParametersMatching: []
      },
```
Committed it now.

image

Tushar-4781:
@tijlleenders where is the **manifest** file

tijlleenders:
> @tijlleenders where is the **manifest** file

It is being generated by the Vite-PWA plugin (hardcoded in the config - could be moved out if you prefer).
I removed the separate file because we should only have the config in one place.

tijlleenders:
image

Works on Vercel now too. It was missing the icons.

@Tushar-4781 Could you finish up the test case fix?
Then I think we can merge.

Tushar-4781 (approved):
this will be the **first biggest achievement** in the **history of the ZinZen App** Cheers 🍻

pulkitvyas08 (approved):
This is amazing work! Installation doesn't work for firefox but does for chromium based browsers so good enough. Approved and merging 🚀

tijlleenders:
Closes #231 #207 #208

PR #417, opened 2022-06-15
Fix: UI Issues in MyFeelings Page

### LINKED ISSUE #411: Show preview of note

### LINKED ISSUE #416: Unable to add Note for the selected Feeling
It seems like the onClick function of feeling(btn) and its Journal (inside it) are overlapping which is changing the state in the wrong order and because of that feeling note is not getting stored correctly into the DB cuz the state here is the id. And for modifying the feeling note it is using the id from the state

Tushar-4781:
Resolves #411 
Resolves #416


Tushar-4781:
> @Tushar-4781 Should the test be failing?

checking

tijlleenders (dismissed):
This fixes the mentioned bug.
Can't judge the React code.

tijlleenders (approved):

pulkitvyas08 (approved):

PR #444, opened 2022-06-20
Implemented: Deisgn Modifications in addFeelingsPage

### LINKED ISSUE #393: Adjust AddFeelings page
There were some changes in the design so the AddFeelings page needs to be modified.
NB: This is not the MyFeelings page, but the page you get after clicking + Feeling.

No longer blocked by #384

Old:
image

New:
image

\+ at bottom of page:
image
@mallikarai05 Personally I don't think this is necessary. If we change our mind (don't want to enter any feeling there is always the back arrow top-left). Do you agree?

Clicking on the left side of a feeling option logs the main feeling (ie Happy, Sad, etc).
Clicking on the arrow gives a popup where you can select 0, 1 or multiple sub-feelings:
image

The custom feeling "I feel ..." does not have to be implemented for this issue to be closed.

Tushar-4781:
Resolves #393

### DIFF
--- a/src/api/FeelingsAPI/index.ts
+++ b/src/api/FeelingsAPI/index.ts
@@ -52,6 +52,24 @@ export const getFeelingsBetweenDates = async (startDate: Date, endDate: Date) =>
   });
 };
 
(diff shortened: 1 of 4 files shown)


pulkitvyas08 (dismissed):
LGTM

Issue #424, opened 2022-06-17
Make the options on my goal items working


pulkitvyas08:
// TODO: Add description

PR #451, opened 2022-06-22
Extend Goals DB

### LINKED ISSUE #425: Extend the database and API of my goals to add sublists
// TODO: Add description

pulkitvyas08:
Resolves #425 
Part of #424, issue to be completed with @Tushar-4781's work on the issue, this PR has no dependency on that

NOTE: This PR will bring breaking changes to the indexdDB storage and interaction. We can discuss the implementation of a migration feature to avoid such changes in the future

### DIFF
--- a/src/models/GoalItem.ts
+++ b/src/models/GoalItem.ts
@@ -7,9 +7,10 @@ export interface GoalItem {
   id?: number;
   title: string;
   duration: Number;
-  sublist: string[] | null;
+  sublist?: GoalItem[];
   repeat: Repeat | string;
   start: Date | null;
   finish: Date | null;
   createdAt: Date;
+  status: 0 | 1; // 0 = active, 1 = archived
 }
--- a/src/models/GoalItem.ts
+++ b/src/models/GoalItem.ts
@@ -4,7 +4,7 @@ export enum Repeat {
 }
 
 export interface GoalItem {
-  id?: number;
+  id?: Number;
   title: string;
   duration: Number;
   sublist?: GoalItem[];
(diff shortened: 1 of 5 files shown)


src/models/GoalItem.ts
```diff
@@ -4,12 +4,14 @@ export enum Repeat {
 }
 
 export interface GoalItem {
-  id?: number;
+  id?: Number;
   title: string;
   duration: Number;
-  sublist: string[] | null;
+  sublist?: GoalItem[];
```
Tushar-4781, line 10: @pulkitvyas08 are you sure it should have GoalItem[] Interface?
@tijlleenders in the sublist is there something other than a string?  Acc to figma board like if I want to go shopping eve1y Wednesday then the sublist will have the items of it right?

Tushar-4781 (changes requested):
LGTM, just wanted confirmation on the below comment. will approve after that

src/models/GoalItem.ts
pulkitvyas08: Yes, refer to the attached snippet. All the sublist items can be deleted, archived, and have their title updated

image

Tushar-4781: cool then. I think we need better example though because scenarios like above i dont see use of other attributes

pulkitvyas08: The only one I see not being used is the plus icon. As sublist items cannot have their own sublist. Rest are good

tijlleenders: The idea is that any list can have a subList, including subLists.
It's a Directed Acyclical Graph hierarchy that can go x levels deep.
That's why I propose separate goals, relationships and properties repositories.

Issue #422, opened 2022-06-17
Dummy archive page (pre-design)


pulkitvyas08:
// TODO: Add description

tijlleenders:
@Tushar-4781 I guess this one is obsolete now that we have suggestions from archive?

Tushar-4781:
@tijlleenders yes, we can close this as we are having both the public goals archive and the personal archive as well.

PR #454, opened 2022-06-22
Feature/goal db save

### LINKED ISSUE #356: Avoid empty 'My goals' page
If someone opens the My goals page without any registered goala it should open the Add goals page instead.

### LINKED ISSUE #452: Store the goal into DB
TODO: In the AddGoals page, the addGoal button should save the goal details into DB.

pulkitvyas08:
Resolves #452 
Resolves #356 

Note regarding #356: at the moment we are redirecting the user to `AddGoals` page even if there are goals in the collection that are archived. This might change if we decide to add the archived goals list to `MyGoals` page itself but since #422 is in the pre-design phase I'm making the assumption that this is the right choice.

### DIFF
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -32,7 +32,11 @@ export const getAllGoals = async () => {
 
 export const getActiveGoals = async () => {
   const activeGoals: GoalItem[] = await db.goalsCollection.where("status").equals(0).toArray();
-  console.log(activeGoals);
+  return activeGoals;
+};
+
+export const getAllArchivedGoals = async () => {
+  const activeGoals: GoalItem[] = await db.goalsCollection.where("status").equals(1).toArray();
   return activeGoals;
 };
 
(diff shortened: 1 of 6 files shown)


tijlleenders:
Separate issue will have yo be made to fix the visualization on MyGoals. Title and color are not showing - and there is a small square that shouldn't be there.

pulkitvyas08:
> Separate issue will have yo be made to fix the visualization on MyGoals. Title and color are not showing - and there is a small square that shouldn't be there.

Uhh this was all working on my local. Let me check again

pulkitvyas08:
> Separate issue will have yo be made to fix the visualization on MyGoals. Title and color are not showing - and there is a small square that shouldn't be there.

Okay I think it would be better to make a new issue to deal with the UI stuff + validation (to not let empty feelings through) as everything else is working just fine

pulkitvyas08:
@Tushar-4781 Can you look at this cypress failure and tell me what's wrong here. Attaching the screenshot of the failure point. I don't think this is the expected behavior here

DashboardUserChoice Test -- Should have My Feelings button working (failed)

Why should `AddFeelings` path contain "MyFeelings"?

Tushar-4781:
This is happening because of that timeout function since Cypress doesn't found anything in db so the app redirect it to addFeelings this is the reason why Cypress tests have been failing from very beginning.

tijlleenders:
@Tushar-4781 can you make an issue to fix this test? Merging as this PR did what it needed to do and the test is a false negative.

PR #459, opened 2022-06-23
Fix: Dashboard Failing TC

### LINKED ISSUE #456: Fix the MyFeelings test in cypress
The test has been failing frequently because of the timeout function in the myFeelingsPage, it redirects to addFeelings. This has to be resolved

Tushar-4781:
Resolves #456

Added a wait command in the test so that cypress will wait while redirect happens


pulkitvyas08 (approved):
The code seems fine. Hope you have run the test to see if everything works fine.

Tushar-4781:
> The code seems fine. Hope you have run the test to see if everything works fine.

Obviously,🤣. It's passing here so double check

PR #466, opened 2022-06-24
Patch/delete sublist with parent

### LINKED ISSUE #465: Sublist goals aren't deleted with the parent goal
This might change in the future and be converted to just removing the link to the parent but for now, it only makes sense to delete the goals in the subgoals when you delete the parent goal

Same should be done for archiving goals

pulkitvyas08:
Resolves #465


Tushar-4781 (approved):
LGTM

PR #483, opened 2022-06-25
Function for populating dummy goals

### LINKED ISSUE #476: Make dummy content for 7 days of tasks
When going to the MyTime page Task collection should be emptied if anything was in it + filled with a dummy set of Tasks with different colors and durations for Today + next 7 days.

Tushar-4781:
Resolves #476

I have kept this function in MyTime Page for now. Since that one is in progress. It would prevent unnecessary navigation while testing. Kindly navigate to MyTimePage to initiate the call. 
PS: Just want to keep development fast XD


tijlleenders:
Navigate back button on subGoal takes me back to Dashboard. Maybe log as separate issue.

tijlleenders (approved):

Issue #482, opened 2022-06-25
[BUG] addGoal btn isn't at bottom right 


Tushar-4781:
On MyGoalsPage, addGoal btn should at bottom right 
![image](https://user-images.githubusercontent.com/39093015/175782264-3701a549-5c47-472c-a19e-bd95b9cf4dc1.png)

nishant-666:
How to get added as a collaborator?

tijlleenders:
> How to get added as a collaborator?

Hi Nishant, thanks for your interest.

You don't have to be an official collaborator to contribute. You can always fork the repo and submit a PR to this repo.

Instructions can be found here: https://docs.github.com/en/get-started/quickstart/contributing-to-projects

I'll email you some info on the project.

nishant-666:
I have opened a pull request. Please check once.

tijlleenders:
Closed by #484

Thanks a lot Nishant!

PR #484, opened 2022-06-25
Add Button postion changed to bottom right.

nishant-666:


tijlleenders (changes requested):
Thanks for this Nishant.

The test case wasn't clearly specified as @Tushar-4781 just made this BUG issue foe himself to work on layer today/tomorrow.

The + button should float above the page. Now if I add a few goals and the list gets longer, the + button scrolls with the page.

I tried on the Vercel preview link in this PR and the same happens.

Do you have a way to fix that?

nishant-666:
Yeah I understood the issue. It should be a fixed button that is always at
the same position even if we scroll. Is that what you are saying?

tijlleenders (approved):
Thanks a lot Nishant!

This closes #482

PR #488, opened 2022-06-26
Integration in MyTimePage: Duration + Width Normalization + DummyTasks (with Timestamp)

### LINKED ISSUE #477: Show day timeline Task summary in MyTime
Based upon the dummy content in #476, show the colored bar  with different colors for each day in the MyTime page.

### LINKED ISSUE #481: Normalize the width of unplanned time in MyTimePage
let's say if on a particular day if there are 6h (5 AM - 11 AM) that are occupied by goals and the rest of 18 hrs are unoccupied then in that case our line would be gray mostly. And because of this, the width of colors related to the duration of tasks will be short because the majority of the time is unplanned.

### LINKED ISSUE #490: Create A toggle button to compare the changes of normalization on MyTimePage
#481 was resolved by #488, a normalization feature was introduced. But to compare the changes which are made by the feature we need to have a toggle button that can flip the changes btw normalized and non-normalized color of widths. the toggle button can help us to test multiple scenarios ( by altering or changing the order of our task durations )

Tushar-4781:
Resolves #477
Resolves #481 
Resolves #490 

Upgrade in solution for #476

To make review easy ( for #481 ) I've made a commit with a message **checkpoint: Preview**. Consider the vercel deployment preview link after this commit to compare and differentiate with the last deployment link ( **checkpoint: Final Preview** )

1. For testing purpose, i have created the function for populating dummy tasks. Creating a collection and set of apis for MyTimePage has to be discussed so till then to keep the code complexity minimum i m using the goals as tasks.
2. The "Unplaned" word represents the user have no tasks in that particular duration i.e. **unplanned time**. Rest other dummy names represent the task name like "Gym" -> "Go to Gym" 
3. Reason for introducing the point 2 is to have a record of the entire day i.e. 24hrs
4. Now since we have the record of all day, there are two things we can find out. **First,** we will know the task that is of the highest duration. **Second,** we can find how much max unplanned time appeared btw two consecutive tasks. 
5. Once we have the MaxUnplannedTime we will take it certain amount and will distribute it among the width of other goals.
**NOTE:** All of the normalization stuff is done to normalize the width of colors (CSS) not the actual duration of tasks.

### DIFF
--- a/src/pages/MyTimePage/MyTimePage.tsx
+++ b/src/pages/MyTimePage/MyTimePage.tsx
@@ -11,6 +11,8 @@ import "./MyTimePage.scss";
 
 export const MyTimePage = () => {
   const [tmpTasks, setTmpTasks] = useState<GoalItem[]>([]);
+  const [goalOfMaxDuration, setGoalOfMaxDuration] = useState(0);
+  const [maxDurationOfUnplanned, setMaxDurationOfUnplanned] = useState(0);
   const today = new Date();
   today.setDate(today.getDate() + 1);
 
@@ -18,6 +20,7 @@ export const MyTimePage = () => {
 
   const getDayComponent = (day: string) => {
     let colorIndex = -1;
+    console.log(goalOfMaxDuration, maxDurationOfUnplanned);
     return (
       <div key={`day-${day}`} className="MyTime_day">
         <div className="MyTime_navRow">
@@ -41,7 +44,7 @@ export const MyTimePage = () => {
               <div
                 key={`task-${task.id}`}
                 style={{
-                  width: `${task.duration * 4.15}%`, // "10%",
+                  width: `${task.duration * 4.17}%`, // "10%",
                   height: "10px",
                   backgroundColor: `${task.title === "Unplaned" ? "gray" : darkrooms[colorIndex]}`
                 }}
@@ -56,15 +59,16 @@ export const MyTimePage = () => {
 
   useEffect(() => {
     (async () => {
-      const random = (min: number, max: number) => Math.floor(Math.random() * (max - min)) + min;
+      // const random = (min: number, max: number) => Math.floor(Math.random() * (max - min)) + min;
       const createDummyGoals = async () => {
         console.log("wait");
         const dummyNames: string[] = ["Unplaned", "Gym", "Study", "Unplaned", "Shopping", "Code Reviews", "Unplaned", "Algo Practice"];
-        dummyNames.map(async (goalName: string) => {
+        const dummyDurations : number[] = [4, 3, 1, 2, 3, 2, 6, 3];
+        dummyNames.map(async (goalName, index) => {
           const dummyGoal = createGoal(
             goalName,
             true,
-            goalName === "Unplaned" ? random(2, 5) : random(1, 3),
+            dummyDurations[index],
             null,
             null,
             0,
@@ -74,12 +78,27 @@ export const MyTimePage = () => {
           return id;
         });
       };
-      let goals: GoalItem[] = await getActiveGoals();
-      if (goals.length === 0) {
+      const getTasks = async () => {
+        const goals: GoalItem[] = await getActiveGoals();
+        let GMD = 0;
+        let MDU = 0;
+        goals.map((goal) => {
+          if (goal.title === "Unplaned" && MDU < goal.duration) {
+            MDU = goal.duration;
+          } else if (GMD < goal.duration) GMD = goal.duration;
+          return null;
+        });
+        setGoalOfMaxDuration(GMD);
+        setMaxDurationOfUnplanned(MDU);
+        return goals;
+      };
+      let tasks: GoalItem[] = await getActiveGoals();
+      if (tasks.length === 0) {
         await createDummyGoals();
-        goals = await getActiveGoals();
       }
-      setTmpTasks([...goals]);
+      tasks = await getTasks();
+
+      setTmpTasks([...tasks]);
     })();
   }, []);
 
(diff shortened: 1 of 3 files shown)


Tushar-4781:
image
image

tijlleenders:
Seems like the Tasks are calculated dynamically based on the Goals?

pulkitvyas08:
I tried the preview, when I have two goals on the list it shows 3 tasks in MyTimePage

image

pulkitvyas08:
> I tried the preview, when I have two goals on the list it shows 3 tasks in MyTimePage
> 
> image

Okay nevermind. This means that there are two tasks + remaining time

pulkitvyas08:
Can we show somewhere what the color means? It's super confusing right now

Tushar-4781:
> Can we show somewhere what the color means? It's super confusing right now

I think it was discussed that the "GRAY" color represents the unplanned time rest other colors represents that task that are there on that day

Tushar-4781:
> > I tried the preview, when I have two goals on the list it shows 3 tasks in MyTimePage
> > image
> 
> Okay nevermind. This means that there are two tasks + remaining time

no you have only 1 task rest is Unplanned time.

tijlleenders:
@Tushar-4781 Why do we have to use the "Unplanned" goal explicitly?

If there is 4h of total goals, why not assume the rest is unplanned?

Also, as Pulkit suggested - maybe we should fix the colors of the goals first (show chosen color on MyGoals) so we can use that in the temporary create Tasks from Goals function?

tijlleenders:
Can we use a special color for unplanned? One that is not part of the regular Goal colors?

Tushar-4781:
> @Tushar-4781 Why do we have to use the "Unplanned" goal explicitly?
> 
> If there is 4h of total goals, why not assume the rest is unplanned?
> 
> Also, as Pulkit suggested - maybe we should fix the colors of the goals first (show chosen color on MyGoals) so we can use that in the temporary create Tasks from Goals function?

MyTimePage is not ready to take user tasks/goals atm. It is designed acc to dummy tasks that are created in that component. We have to format our data in a structured way that is reqd for this page.

tijlleenders:
> Can we show somewhere what the color means? It's super confusing right now

Yeah that will be done in a separate issue. You can then uncollapse the summary to see the day schesule in detail.

pulkitvyas08:
> 

Format the data as in changing the Goals collection schema or format the data on the page?

Tushar-4781:
Point to be noted: the whole line of colors represents the clock i.e. 24 hrs. any color apart from "gray" represents that it is task otherwise its an unplanned time

Tushar-4781:
> > 
> 
> Format the data as in changing the Goals collection schema or format the data on the page?

format the data on page. I m not changing the Goal collection schema here.

pulkitvyas08:
> Point to be noted: the whole line of colors represents the clock i.e. 24 hrs. any color apart from "gray" represents that it is task otherwise its an unplanned time

If that's the case then can you add a few dummy goals here with `start` and `end` times included that would show gaps between the tasks and validate what you are saying

Tushar-4781:
@pulkitvyas08  @tijlleenders if it still confusing to differentiate between tasks. i can make that uncollapse feature right away in this PR itself. it won't take much time

Tushar-4781:
> > Point to be noted: the whole line of colors represents the clock i.e. 24 hrs. any color apart from "gray" represents that it is task otherwise its an unplanned time
> 
> If that's the case then can you add a few dummy goals here with `start` and `end` times included that would show gaps between the tasks and validate what you are saying

yeah I can do that as well.

Tushar-4781:
> @Tushar-4781 Why do we have to use the "Unplanned" goal explicitly?
> 
> If there is 4h of total goals, why not assume the rest is unplanned?
> 
> Also, as Pulkit suggested - maybe we should fix the colors of the goals first (show chosen color on MyGoals) so we can use that in the temporary create Tasks from Goals function?

@tijlleenders yeah the unplanned should not be in the DB, it's confusing others. it should be made while we get all the tasks only. I will make the changes to resolve the above issues mentioned by u guys.

Tushar-4781:
Resolves #
1. There is a toggle button called "normalize" you can see how it makes a difference between colors
2. You guys can also click on the arrow to check the tasks based on which the colors are decided

NOTE : 
- light Gray color denotes that this time is unplanned
- the whole line of colors == 24hrs so assume that start if from 00:00 - 23:60

tijlleenders:
Nice work.

I think this solves what the PR was set to do - and then some!

tijlleenders (approved):

pulkitvyas08:
> 

I don't think this was done

pulkitvyas08:
> Resolves #490
> 
>     1. There is a toggle button called "normalize" you can see how it makes a difference between colors

Does this still stand? I can't see any button on the preview

src/pages/MyTimePage/MyTimePage.tsx
```diff
@@ -8,65 +8,164 @@ import { addGoal, createGoal, getActiveGoals } from "@src/api/GoalsAPI";
...
+          }
+          if (GMD < goal.duration) GMD = goal.duration;
+          return null;
+        });
+        unplannedInd[0] = unplannedInd[0] === -1 ? 0 : unplannedInd[0];
+        setUnplannedDurations([...unplannedDur]);
+        setUnplannedIndices([...unplannedInd]);
+        setGoalOfMaxDuration(GMD);
+        setMaxDurationOfUnplanned(MDU);
+        console.log(unplannedInd,unplannedDur);
```
pulkitvyas08, line 159: Remove this

pulkitvyas08 (approved):
Code-wise this is okay. Good work, remove the console log and we are gtg

Tushar-4781:
> > Resolves #490
> > 
> >     1. There is a toggle button called "normalize" you can see how it makes a difference between colors
> 
> Does this still stand? I can't see any button on the preview
> 

@pulkitvyas08 it is in the same line of "My Time" title

Tushar-4781:
> > 
> 
> I don't think this was done

@pulkitvyas08 What's missing ?

tijlleenders (approved):
Some tough work!

PR #497, opened 2022-06-28
Patch/optional goal magic

### LINKED ISSUE #467: Make goal magic optional
Right now it is compulsory to add "1h" and "daily" to a goal.
It should be optional.

So "Walk" and "Walk 1h" are valid goals and should be stored.
It is not compulsory to enter "Walk 1h daily".

pulkitvyas08:
Resolves #467


Tushar-4781 (approved):
LGTM

PR #503, opened 2022-06-28
Update/subgoal color

### LINKED ISSUE #500: Show Goal color for subGoals too
We now show the custom Goal color ... but it doesn't work for subGoals yet.

Technically in the previous version of the app I used to not make a difference between Goals and subGoals. It's just a oage where you render the list + the corresponding breadcrumb. Maybe that's something that can be refined here too. Would probably require to be broken down into several issues.

...or just add it to the subGoals page!

pulkitvyas08:
Resolves #500


tijlleenders (approved):
That was quick!

PR #506, opened 2022-06-29
Show goal options on arrow tap, navigate on click on the goal

### LINKED ISSUE #505: Change what clicking on a goal does
Right now:
Click on a Goal opens sub icons.
Click on > goes to subGoal.

Wanted:
Click on a Goal goes to subGoal.
Click on > opens sub icons.

pulkitvyas08:
Resolves #505

### DIFF
--- a/src/pages/MyGoalsPage/MyGoalsPage.scss
+++ b/src/pages/MyGoalsPage/MyGoalsPage.scss
@@ -86,7 +86,6 @@
   font-weight: bold;
   font-size: 1.3em;
   text-align: left !important;
-  cursor: pointer;
 }
 @media (min-width: 1281px) {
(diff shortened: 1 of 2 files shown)


tijlleenders (changes requested):
Looks good.
This can close the issue.

There is a regression effect:
The delete icon doesn't work any more but instead collapses the sub icons.

pulkitvyas08:
> Looks good. This can close the issue.
> 
> There is a regression effect: The delete icon doesn't work any more but instead collapses the sub icons. The add subGoal icon too.

I will do, we definitely need tests to avoid these situations 😅

tijlleenders (approved):
Thanks!

PR #511, opened 2022-07-02
Added the back to previous page functionality to the arrow

GithinjiHans:


pulkitvyas08 (dismissed):
Pretty simple fix and works as intended. But not sure if this is the expected behavior. Let's take this case for example:

First-time user, no goals in the list. I go to `/AddGoals` and create a goal so it sends me to `/MyGoals` then I click on the goal to go to `/MyGoals/GoalId` to view the sublist. Then I click on add subgoal button which redirects to `/AddGoals`, and finally adding a subgoal sends me back to `/MyGoals/GoalId`

Now the history navigation stack is:
`Home` -> `/AddGoals/` -> `/MyGoals ` -> /MyGoals/GoalId` -> `/AddGoals` -> `/MyGoals/GoalId`

Clicking on the back button will send us through this stack to get back to the home page

My questions is that shouldn't this stack be:
`Home` -> `/MyGoals` -> `MyGoals/GoalId`

Do we really want to go to the `/AddGoals` page again and go to the same page twice?

@tijlleenders Your thoughts?

pulkitvyas08:
If this is indeed the expected behavior then @GithinjiHans just fix the test errors and we are good to go

tijlleenders:
I agree this needs some more refinement.

This PR can go through. Let's use it for a while and see.

I'll make an issue for filtering pages like AddGoal from the navigation stack, that should also apply to built-in Android back button.

pulkitvyas08 (approved):

PR #512, opened 2022-07-02
Patch: MyTimePage UI/UX 

### LINKED ISSUE #493: Change naming of days in My time
The screen looks cluttered with all the dates.

Can we only use the full date for "Today, Sun 26 June"

Then use "Tomorrow", "Tuesday", "Wednesday", etc?

### LINKED ISSUE #501: Uncollapse day timeline by click on block
Right now we have to specifically click the > to uncollapse a day. It would be better if you can click anywhere on the day block to uncollapse the day.

To collapse click anywhere on the block too - unless the tap is 'caught' by a Task.

### LINKED ISSUE #502: (un)collapse day blocks independently
Right now, when uncollapsing a day block all other day blocks are automatically closed.

Collapsing blocks above the one you're uncollapsing automatically makes the page jump in unexpected ways.

Wanted:
It would be better if blocks that are uncollapsed remain uncollapsed until closed. 

That way you can compare details of the days and browse through the week without having to uncollapse every day one by one.

Tushar-4781:
Resolves #493 
Resolves #501 
Resolves #502

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -4,6 +4,8 @@ export const formatDate = () => {
   return newDate;
 };
 
+export const weekday = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"];
+
 export const getJustDate = (fullDate: Date) => new Date(fullDate.toDateString());
 
 export const truncateContent = (content: string, maxLength = 20) => {
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -4,8 +4,6 @@ export const formatDate = () => {
   return newDate;
 };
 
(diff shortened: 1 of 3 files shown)


src/utils/index.ts
```diff
@@ -4,6 +4,8 @@ export const formatDate = () => {
   return newDate;
 };
 
+export const weekday = ["Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"];
+
```
tijlleenders, line 8: Isn't there a time library already loaded that can do this?

tijlleenders (commented):
Can you remove "My time" and replace "Today" by "My Sat 2 Jul 2022"?

tijlleenders (approved):

PR #522, opened 2022-07-03
assets should get served from cache

GithinjiHans:


src/service-worker/service-worker.ts
```diff
@@ -13,7 +13,7 @@ import { clientsClaim } from "workbox-core";
 import { ExpirationPlugin } from "workbox-expiration";
 import { precacheAndRoute, createHandlerBoundToURL } from "workbox-precaching";
 import { registerRoute } from "workbox-routing";
-import { StaleWhileRevalidate } from "workbox-strategies";
+import { CacheFirst } from "workbox-strategies";
```
tijlleenders, line 16: Why are we changing strategies?

GithinjiHans: Assets are served from the cache storage using CacheFirst strategy....if they are not found in the cache, they are requested from the server then cached and the response from the server is served initially.

tijlleenders: Thanks for the explanation - but how is this different from stale-while-revalidate?

pulkitvyas08: While the implementation details under the hood are quite similar for both, what I observed is that the fonts and SVG are loading much faster in this version as compared to the zinzen.vercel.app. @tijlleenders Can you also do the same comparison on, say regular 2G, and check if that's true

tijlleenders: @GithinjiHans Can you tell me what the theoretical difference is between the two caching strategies CacheFirst and StaleWhileRevalidate?

tijlleenders (changes requested):
The font is not cached yet.

src/service-worker/service-worker.ts
GithinjiHans: Stalewhile-revalidate also serves assets from cache but the assets in the cache are checked if they're upto date, if not they are not upto date they're updated and the latest version is served mostly used in live apps. Thats the main difference between Stalewhile-revalidate and Cache first.

GithinjiHans:
> The font is not cached yet.



> The font is not cached yet.

I am unable to produce the error on FF I'll correct it once am able to produce it on chrome

src/service-worker/service-worker.ts
tijlleenders: Ok - so we want [stale-while-revalidate](https://developer.chrome.com/docs/workbox/modules/workbox-strategies/#stale-while-revalidate) - not [cache-first](https://developer.chrome.com/docs/workbox/modules/workbox-strategies/#cache-first-cache-falling-back-to-network) as cache first will never update the cache.

If the cache is never updated then users will never get updates.
So unless you have an option to set an interval for checking/updating the cache we should use stale-while-revalidate.

tijlleenders:
> > The font is not cached yet.
> 
> > The font is not cached yet.
> 
> I am unable to produce the error on FF I'll correct it once am able to produce it on chrome

Hint: the font is a .woff2 file and you have not configured anything to cache that - so it should be reproducible.
image

src/service-worker/service-worker.ts
tijlleenders: Page loads using stale-while-revalidate should be just as fast as page loads using cache-first, right?

tijlleenders, line 16:
```suggestion
import { StaleWhileRevalidate} from "workbox-strategies";
```

tijlleenders:
Even though CacheFirst is not faster than StaleWhileRevalidate it is preferred because it doesn't cause unnecessary network traffic, so it is:
- Cheaper
- Safer (no usage info 'leaking' to the network)

We don't have to build anything to refresh the cache: When a file changes the content hash in the filename will change => forcing a cache miss and load from network.

@GithinjiHans Thanks for changing it!

tijlleenders (dismissed):

tijlleenders (approved):

PR #525, opened 2022-07-05
Fixed the text overflow error in my goals page

GithinjiHans:


tijlleenders:
It is not exactly what was asked since 'ellipsed' means cutting off a title and adding "...".

However - this also solves the problem and might even be better since you can then view the whole title.

image

PR #526, opened 2022-07-05
Added  ® to all Zinzen occurences where neccessary.

GithinjiHans:


tijlleenders:
@GithinjiHans Can you also do the readme?

tijlleenders:
It is actually pretty ugly to have the big R everywhere, especially with this font. 
Can you try to make it smaller, less obvious?

GithinjiHans:
> It is actually pretty ugly to have the big R everywhere, especially with this font. Can you try to make it smaller, less obvious?

Or maybe make it superscript if possible

mallikarai05:
@tijlleenders Sure, I'll try something out

tijlleenders:
Size is a lot better.
Positioning?

GithinjiHans:
> Size is a lot better. Positioning?
> 
> 

We can only design the R individually and the multilanguage feature makes it even more complicated because the position of the word ZinZen is different. For now we can just use the R in the userchoice dashboard and delete the others in other pages then rethink of how to go about that.

tijlleenders:
Ok good work.
Indeed we should not invest too much in making this possible. 

Removing the occurences by rephrasing good be an option, for example:
- "Quick intro"
- "Skip intro"

GithinjiHans:
> Ok good work. Indeed we should not invest too much in making this possible.
> 
> Removing the occurences by rephrasing good be an option, for example:
> 
>     * "Quick intro"
> 
>     * "Skip intro"

Good idea...so we should use this PR for that..right??

tijlleenders:
Sure!

tijlleenders:
@Tushar-4781 Can you assist with the failing  test (see last 3 commits)?

Tushar-4781 (commented):
The tests are failing because of the update "Skip intro". We have to update all the tests not just the dashboard,

tijlleenders (approved):

PR #529, opened 2022-07-09
Made my time page the default entry page

GithinjiHans:


Tushar-4781 (commented):
since the dashboard is no longer the default entry page. starting test of all other pages has to be modified.

tijlleenders (changes requested):
@GithinjiHans 

The first time I open the app it works as expected. I go through the setup choices (language and theme) and intro, thrn I land on MyTime.

However as a returning user, the app still opens on the Dashboard.

Can you fix that?

GithinjiHans:
> @GithinjiHans
> 
> The first time I open the app it works as expected. I go through the setup choices (language and theme) and intro, thrn I land on MyTime.
> 
> However as a returning user, the app still opens on the Dashboard.
> 
> Can you fix that?
Sure

tijlleenders (approved):

PR #545, opened 2022-07-16
Modified : Approach for Navigation in MyGoals & AddGoals Page 

### LINKED ISSUE #539: Make a single MyGoals page capable of also navigating to sub(sub...)Goals
Right now we have a page for
- MyGoals (entry page for top level goals)
and
- MySubGoals (only for viewing the n-1 level of the goal hierarchy)

Wanted:
Merge the two pages so that navigation between goals and sub-goals stays on the same page.

Tushar-4781:
Resolves #539


tijlleenders (approved):
Great!
Not leaving the page while navigating will improve user experience a lot!

Issue #361, opened 2022-05-22
How to integrate the wasm in the app CI?


tijlleenders:
This issue is foe the diacussion on how to integrate the wasm module to the frontend.

The wasm is being built in a different repo.
We need a place where the app CI can pull the latest version of the WASM module.

@sokorototo For now we can copy manually when a new version is available.

See comment https://github.com/tijlleenders/ZinZen/issues/361#issuecomment-1193251470

sokorototo:
We probably need to get git submodules working

tijlleenders:
Submodules doesn't look like a good fit.
Reasons on [this GitHub blogpost](https://github.blog/2016-02-01-working-with-submodules/)

What about semantic versioning on crates.io and then pulling it from there?

sokorototo:
You can't pull binaries from crates.io and publishing a module to npm is overkill. Instead one can just use Github's CDN based on repos. Eg the link to the raw wasm file would be: <https://github.com/tijlleenders/ZinZen-scheduler/blob/main/ts/scheduler.wasm?raw=true>

tijlleenders:
Yeah, if that works it is a lot simpler!

sokorototo:
A problem is that always links to the latest development version, we can't just ship code in testing to users. We have to make a release branch then link the wasm from there

tijlleenders:
For now it's important to get a workinf integration. So copy the files manually for each new wasm release. 

Then we can think about automating the integration.

The files can be generated/integrated in several ways. @xy2iii gave us a nice explanation in https://github.com/tijlleenders/ZinZen-scheduler/issues/24

There are a few experimental Vite plugins for output 'bundler'. If that doesn't work we can switch to web, nodejs or no-bundler if that does work (https://rustwasm.github.io/wasm-bindgen/reference/deployment.html).

sokorototo:
Do git submodules work with CI? Because we can just link the js-api folder during building, and webpack takes it from there

tijlleenders:
GitHub is [quite negative](https://github.blog/2016-02-01-working-with-submodules/)on using Git submodules - it's a "last resort". This was in 2016 so might have improved since.

My first impression is that a git submodule would not be a good choice because:
- incorporate the whole ZZ-scheduler repo as a subdirectory - not just the js-api dir
- you have to issue a command to update it - so then we might as well have a script that doeals a raw download of the required files from zz-scheduker into the ZZ repo and adds a commit

[...]

PR #548, opened 2022-07-17
533 call the scheduler on mytime page

GithinjiHans:


tijlleenders (changes requested):
My bad.
A discussion is needed before continuing on this PR. There is a discussiom somewhere with Newton and/or @pulkitvyas08 that using git submodules is not the way to go.
I should have included more information.

GithinjiHans:
Yeah...They are not actually important right now. I can remove the submodule and It will still work.

tijlleenders:
I was assuming you would discuss this with @sokorototo - as you see him frequently and he's doing the WASM part.
The discussion where we decide not to use submodules is in #361

tijlleenders:
> Yeah...They are not actually important right now. I can remove the submodule and It will still work.

Sure, thanks for taking it that way :)
From #361 I deduce we don't want the wasm in /src
Not sure if the same is true for the api.ts - we should have it in one place, not in two repos.

GithinjiHans:
Sure...we don't need the wasm because it's loaded using a URL but I think the api.ts is needed.

GithinjiHans:
I've tried using the URL to access the api.ts and it just fails.
We can draft this PR and discuss on where to keep the api.ts ....backend or frontend?

tijlleenders:
This PR was for trying to integrate by caloing the wasm module via the GH URL from the frontend code.

That didn't work so we'll start fresh in a new PR for the other methods discussed in #361

PR #550, opened 2022-07-18
Re-Development: MyGoalsPage & Fix: of Navigation between pages and their components

### LINKED ISSUE #514: Navigate with replace after adding Goal/Feeling
https://reactrouter.com/docs/en/v6/hooks/use-navigate

When navigating away from AddFeeling or AddGoal, use replace = true so that when user presses back button - you don't go back to the addFeeling/Goal screen.

### LINKED ISSUE #547: Update MyGoals Page to have N levels of subgoals
Currently, we are following 2 level structure i.e. one goal and then its subgoal. But now it's time we allow to have goals under a subgoal i.e. sub(sub)goal.

### LINKED ISSUE #549: Re-develop the Header for MyGoalsPage
The current approach of using header ain't good. It's a common component that should not be used in components of other pages. The goal is to prevent reloads and stop navigation via URL for that the current approach is not suitable, it has to be modified for myGoalsPage.

Tushar-4781:
Resolves #514 
Resolves #547 
Resolves #549


pulkitvyas08:
@Tushar-4781 AddGoals page is giving 404. Any particular reason?

Tushar-4781:
> @Tushar-4781 AddGoals page is giving 404. Any particular reason?

@pulkitvyas08 the commit is on the way. its taking time in syncing the changes T_T

GithinjiHans (approved):
Everything's good on my side.

pulkitvyas08:
Not sure if this is in the scope of this PR but adding a goal doesn't immediately update the goals list on MyGoals page. Check the attached video for reference

https://user-images.githubusercontent.com/54947769/179744640-05ca919c-a013-4c9a-a9c8-d831f1db6b21.mp4

pulkitvyas08:
Again, I'm not sure if this comes under the scope of this PR but a couple of things:
- Breadcrumbs are not working on sub-goal pages. Like clicking on the first breadcrumb - "My Goals" takes me to the MyGoals page, that's fine, but the other breadcrumbs don't work
- Deleting a top-level goal doesn't delete all the sub-goals under it. Is this expected behavior? Because the thing is if you delete a parent goal then there is no way to access children's goals, it's just in the database and not linked to the UI in any way.
- Sub goals don't inherit the color and other properties of parent goals

pulkitvyas08:
> Again, I'm not sure if this comes under the scope of this PR but a couple of things:
> 
>     * Breadcrumbs are not working on sub-goal pages. Like clicking on the first breadcrumb - "My Goals" takes me to the MyGoals page, that's fine, but the other breadcrumbs don't work
> 
>     * Deleting a top-level goal doesn't delete all the sub-goals under it. Is this expected behavior? Because the thing is if you delete a parent goal then there is no way to access children's goals, it's just in the database and not linked to the UI in any way.
> 
>     * Sub goals don't inherit the color and other properties of parent goals

Let me know if these will be fixed in some other PR/issue and I'll review your code next and overlook these

Tushar-4781:
> Again, I'm not sure if this comes under the scope of this PR but a couple of things:
> 
> * Breadcrumbs are not working on sub-goal pages. Like clicking on the first breadcrumb - "My Goals" takes me to the MyGoals page, that's fine, but the other breadcrumbs don't work
> * Deleting a top-level goal doesn't delete all the sub-goals under it. Is this expected behavior? Because the thing is if you delete a parent goal then there is no way to access children's goals, it's just in the database and not linked to the UI in any way.
> * Sub goals don't inherit the color and other properties of parent goals

@pulkitvyas08 i was hoping that with this PR we can review the approach of navigation that we are using here. if everyone is satisfied then we can use the same for breadcrumbs ( which will be done in this PR )

As far as  "**Deleting a top-level goal doesn't delete all the sub-goals**" issue is concerned. I didn't modified that implementation its there as it is. It is working for active goals ( i.e. listed on main myGoalsPage ) but not working for their sublist. I will check & modify the implementation for the same. I saw it was implemented in GoalsSublist component. So i assumed it must be working. Good catch. I will fix this in this PR only.

Coming on to "**Sub goals don't inherit the color and other properties of parent goals.**" As per the Figma design, we are providing the color option to users so I thought they can change the color of their subgoals so the code is in sync with that. also in the design samples, I don't see inheriting the same color. If it's about timing & duration, that will be taken care of once we are ready to integrate the scheduler. what other properties do you feel should be in sync? can you specify them, please?

Tushar-4781:
> 

@pulkitvyas08 added the navigation via breadcrumbs. And deleting the goals till n-level is no big deal. I just added a recursive call of 3 lines in your function. 
Leaving the last issue of inheritance for now

tijlleenders:
Nice work @Tushar-4781

PR #564, opened 2022-07-23
Made the goal parser case insensitive

GithinjiHans:


tijlleenders (changes requested):
Can you reproduce this?
The code seems OK at first glance but the preview doesn't work when I remive daiky and type it again with some capitals:

src/components/GoalsComponents/AddGoal/AddGoalForm.tsx
```diff
@@ -42,15 +42,22 @@ export const AddGoalForm: React.FC<AddGoalFormProps> = ({ goalId, setShowAddGoal
       [e.target.name]: value,
     });
   };
-  const freqRegex = /[D|d]+aily|AILY/;
-  const freq = formInputData.inputGoal.match(freqRegex);
+  const daily = /daily/i;
+  const once = /once/i;
+  const weekly = /weekly/i;
+  const freqDaily = formInputData.inputGoal.match(daily);
```
Tushar-4781, line 48: can't we  just lowercase all entire inputGoal while checking. this way we can dodge the case-sensitive issue

tijlleenders: The /i approach has the advantage that it is simple. It might be harder to read for a dev not familiar with the regex options so a comment would be good.

I believe /i doesn't remove the formatting for the matched text? Anyway that is not important for this as we choose/hardcode the case of the property title in the code.

tijlleenders: If the issue I mentioned can be reproduced and it is not clear why the regex fails then Tushar's approach is a good workaround to not spend too much time on this.

GithinjiHans: Actually, I think we should go with Tushar's approach.

tijlleenders:
Can you reproduce my failing test "daiLy"?
Maybe good to add that to the test cases?

GithinjiHans:
> Can you reproduce my failing test "daiLy"? Maybe good to add that to the test cases?

I was able to reproduce it and after using Tushar's approach I was able to solve it

tijlleenders:
Did you add a teat for it?
It still fails when I use the preview link:

Tushar-4781 (commented):
image

image

tijlleenders (approved):
@Tushar-4781 can you comment on the images you posted?
Do you mean to say that the PR can't be approved because the www.x.com is not recognized?
I agree that www.x.etc should be recognized, but I think that should be a separate issue - this PR is just about making the parser case-insensitive.

Tushar-4781:
Yeah that www.x.com ain't working. Let's take care of this in new PR.

Tushar-4781 (approved):

PR #567, opened 2022-07-28
Patch: UX of interacting with Goal Color

### LINKED ISSUE #544: Cycle through colors with 'color order' on AddGoal
Wanted:
When cycling through the colors on the AddGoal the colors should be grouped together. So all the greens, then all the blues, etc. This way it becomes a little easier to find a shade of 'green/blue/etc' you're looking for.

### LINKED ISSUE #562: Update Goal color via edit Goal
Wanted: change color in edit Goal is reflectesd in MyGoals immediately (also in breadcrumb)

Tushar-4781:
Resolves #544 
Resolves #562

### DIFF
--- a/src/components/GoalsComponents/UpdateGoal/UpdateGoal.tsx
+++ b/src/components/GoalsComponents/UpdateGoal/UpdateGoal.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { Container, Row } from "react-bootstrap";
 import { useTranslation } from "react-i18next";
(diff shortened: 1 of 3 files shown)


GithinjiHans (commented):
I think #544 requires that similar colors should be grouped together. Otherwise the color update LGTM.

Tushar-4781:
> I think #544 requires that similar colors should be grouped together. Otherwise the color update LGTM.

yes @GithinjiHans I have placed similar colors adjacent to each other. you can check by clicking on the brush icon. if you feel like there is a better way of arranging. plz feel free to do in this PR

tijlleenders (approved):

PR #575, opened 2022-07-30
Language Tag Visible while adding/updating goal

### LINKED ISSUE #227: Language tag when adding goal
When adding a goal on AddGoal page, the language tag should appear below the title. 

For example, if the user interface is in English, the tag "English" should appear next to the other possible tags like "1h" and "daily".

Tushar-4781:
Resolves #227

### DIFF
--- a/cypress/integration/Feelings.spec.ts
+++ b/cypress/integration/Feelings.spec.ts
@@ -1,5 +1,6 @@
 describe("Desktop Mode Tests", () => {
   it("should have 6 categories of feelings on addFeelings Page", () => {
+    indexedDB.deleteDatabase("ZinZenDB");
     cy.visit("/");
     cy.get(".lang-btn1").contains("English").click();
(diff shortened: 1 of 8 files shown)


tijlleenders (changes requested):
The default goals are English as expected.
However if one chooses a different language at the beginning ie Hindi, I'd expect that language to show - and being saved - not English.

tijlleenders:
When I make a Goal in French mode, the French Goal has the right tag:


... but when I edit it again it has the wrong tag:

tijlleenders (changes requested):
Editing a saved non-English Goal shows the wrong tag.

Tushar-4781:
Sorry @tijlleenders actually the commit is taking time. I think husky is checking or something idk. Will mark this pr draft. Will change it for review as soon as its pushed.

tijlleenders (approved):

Issue #513, opened 2022-07-02
Make Goals with URL property open URL


tijlleenders:
Whenever a Goal has a Link/URL property and it gets clicked in MyGoals it should copy to clipboard or open as shown in roadmap:

![Screenshot_20220702-135130_DuckDuckGo.jpg](https://user-images.githubusercontent.com/44887755/176999611-81c5629c-bbd7-4e13-8ae0-71e9137240d3.jpg)

Blocked by #433 and #434

tijlleenders:
@Tushar-4781 I think this can be worked on as design and link property are done  now:
![Screenshot_20220724-075711_Chrome.jpg](https://user-images.githubusercontent.com/44887755/180634475-186be730-ac07-405e-aff1-0401d5766871.jpg)

If you want to implement a simpler/different design that is also fine - as long as we get a choice to copy or goto the URL/link.

tijlleenders:
#434 is not really blocking as Link property can also be removed by removing the link text.

Tushar-4781:
Blocked by #576

PR #579, opened 2022-08-02
Updated: Goal parser | Added: Link property in GoalItem Interface

### LINKED ISSUE #576: Update the Goal Parser approach
Currently, goalParser that is integrated in **AddGoal** & **UpdateGoal** only detects if duration/link/repetition. We need the starting & ending position of these entities. Current modification in the approach is requested because the URL is part of the goal title and since the URL is part of the goal title, the title becomes long and looks messy on the myGoals page. Instead, we have to replace the url with the text "**_URL_**" which is clickable. So for that, we need to make the modifications such that we can extract these properties along with the help of the positions. Also, a little bit of code cleaning & optimization in terms of memory can be done in the same.

Tushar-4781:
Resolves #576 

The aim of this PR was to modify the goal parser and introduce a new property **link** in Goals schema which will be used to store the link. Earlier URL was part of the title.

The URL that which user enters in the input field will be stored separately. Which means it won't be visible with the goal title.
Instead of the whole link **"URL"** text will appear at the end of the goal title on the MyGoals page and it will be clickable. We have a separate issue for this functionality ( #513 ) and will be resolved a new PR.

### DIFF
--- a/src/__tests__/GoalParser.test.ts
+++ b/src/__tests__/GoalParser.test.ts
@@ -15,7 +15,7 @@ const urlData = (data) => [...data, goalUrl];
 const freqRegex = /daily/;
 const lowercaseInput = goalObject.title.toLowerCase();
 const freq = lowercaseInput.match(freqRegex);
-const urlDetector = /(http|ftp|https):\/\/([\w_-]+(?:(?:\.[\w_-]+)+))([\w.,@?^=%&:\\/~+#-]*[\w@?^=%&\\/~+#-])/;
+const urlDetector = /(?:(?:https?|ftp|file):\/\/|www\.|ftp\.)(?:\([-A-Z0-9+&@#\/%=~_|$?!:,.]*\)|[-A-Z0-9+&@#\/%=~_|$?!:,.])*(?:\([-A-Z0-9+&@#\/%=~_|$?!:,.]*\)|[A-Z0-9+&@#\/%=~_|$])/igm;
 const tracker = /(1[0-9]|2[0-4]|[1-9])+h/i;
 const checkGoal = parseInt(goalObject.title.match(tracker), 10);
 const parseGoal = parseInt(goalObject.title.match(tracker), 10) <= 24;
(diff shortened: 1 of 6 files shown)


tijlleenders:
If I only add a link it doesn't open, and when I edit I see two links:

tijlleenders:
I guess not opening a browser page when clicking on the Goal is OK, as that is not the scope of the PR.

tijlleenders:
I would also expect it to parse www.x.tld - and to be case insensitive (Https://...)

GithinjiHans (commented):
Everything's good I guess

tijlleenders:
Thanks Hanson!
Please coordinate with Tushar who is workimg on this PR though. It would be a shame to both do the same work.

The PR description says:
> The URL that which user enters in the input field will be stored separately. Which means it won't be visible with the goal title.

So when I edit the goal again I expect the URL to be in the title once. However it appears twice. So it looks like it doesn't get removes from the title on save/update but it does get added back on edit, resulting in two urls in the title:

Tushar-4781:
@tijlleenders yes, ig it's happening becasue of two spaces and deviating from the rule which UI as mentioned above. ** task + duration + frequency + url **

tijlleenders:
Ok, then I've not communicated the goal of the parser well. 

There is no fixed structure "title duration etc" Is just as valid as "etc duration title" and should result in the same magic being recognized - except in cases where there are multiple matches possible on a property.

Also every magic property (duration, repetition, URL) is optional.

Finally, there can be only one value for each property. If there are multiple regex matches on a property it's probably best to use the last one.

Take the example "watch show 24h in paradise daily motion weekly 1h URL" - for the property duration you want to match 1h, not 24h. This is assuming you start typing the title and magic is always added last.

tijlleenders (approved):
Yup, this works.

The regex for URLs is a little eager though. Www.x is not a valid URL but gets detected as one...

Also noticed that when we have multiple URLs in the title all but one get 'thrown out'. Fine for now. This was not in scope of the PR and is undefined space.

Nice job!

PR #585, opened 2022-08-10
Integrated: Goal Timing ( Start & Deadline ) magic 

### LINKED ISSUE #557: start magic
Wanted:
When I type "goalTitle 1h start @22" the goal property 'start' should be set to the datetime of today @ 22:00

### LINKED ISSUE #558: deadline magic
Wanted:
When I type "goalTitle 1h before @22" I expect the 'deadline' property of the goal to be set to the datetime of today @ 22:00.

### LINKED ISSUE #584: "Start-deadline" magic
If we type "goalTitle 5-8" it should understand that start is 5:00 and deadline is 8:00.
 However if duration is specified then it should be given the priority i.e. "goalTitle 5-8 1hrs" extract the start and deadline as 5 and 8 but the duration of this task will be 1 hr. So basically It can be done from 5 to 6 or 6 to 7 also.

Tushar-4781:
Resolves #557
Resolves #558
Resolves #584

### DIFF
--- a/src/components/GoalsComponents/AddGoal/AddGoalForm.scss
+++ b/src/components/GoalsComponents/AddGoal/AddGoalForm.scss
@@ -99,7 +99,6 @@ body {
 }
 
 .tags {
-  display: flex;
   font-size: small;
   padding-right: 0.2em;
   border-radius: 1.5em;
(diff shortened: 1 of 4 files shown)


tijlleenders (changes requested):
It works to store, but the text stays in the title so when I 
- enter "Do this 10-11"
- save
- edit saved Goal again

I see this:

tijlleenders (approved):
Perfect for the test case!

PR #602, opened 2022-08-18
Feature added: Add/Remove tag in add/Update Goals page

### LINKED ISSUE #434: Remove/Add URL magic in title by clicking on tag
When a "Link" suggestion is shown as in #433:
A click on the activated Link tag removes the Link property from the Goal but leaves the (now non-magic) text.

A click on the deactivated Link tag adds the Link property to the Goal and converts the URL to magic text.

Clicking on magic URL text in MyGoals is out of scope of this issue.

Tushar-4781:
Resolves #434


tijlleenders (approved):
Awesome!

PR #610, opened 2022-08-19
Update: Suggestions in Public Goals tab

### LINKED ISSUE #601: Show Public Goals in suggestions
Get the Public Goals for the parent goal we're currently in from the internet. So if I'm in 'Shopping list' I want to see what other people have put under their goal 'Shopping list' and shared anonymously.

Tushar-4781:
Resolves #601


tijlleenders (approved):
Yeah!
Another step towards collective intelligence.

PR #614, opened 2022-08-21
Change position of the update task button

GithinjiHans:


tijlleenders:
Overlap on DuckDuckGo mobile.
Please check if also happens on FF/Chrome.

GithinjiHans:
I can't produce it on FF but I guess I know where the problem is

tijlleenders:
Overlap in Chrome too.

PR #625, opened 2022-09-05
Upadated: TextParser in Add/Update Goal

### LINKED ISSUE #623: Update Goal Parser to recognize additional patterns mentioned in Wiki

Tushar-4781:
Resolves #623


tijlleenders (changes requested):
Great!
- Can you use 25h notation hh:mm (so without AM/PM and without milliseconds)?

Other than that good to go!

tijlleenders:
Nice, that's what I meant.
However, the start hour should be 0:00 in thia case. For due it is OK.

tijlleenders (approved):
Perfect!

PR #639, opened 2022-09-10
Patch: triming the spaces from string values in add/update goals

### LINKED ISSUE #638: Improve matching on parent by trimming multiple and ending spaces in titles
Matching happens on hash of title so upon creation/update of a Goal:
- trailing spaces in title should be removed 
- multiple spaces should be reduced to one

Tushar-4781:
Resolves #638


tijlleenders (approved):
Lightning speed!

PR #661, opened 2022-09-28
Update: navigate & show input in goal sublist if add a subgoal choice is made

### LINKED ISSUE #658: Go into goal when adding subgoal from sub icon +
When I add a sublist to for example Shopping using the + on the sub icons of Shopping, I expect to navigate into Shopping first, then add new goal like so:

Instead I see this:

Tushar-4781:
Resolves #658


tijlleenders:
ESLint is still failing?

PR #668, opened 2022-09-28
Update: Added change goal color btn below add/update goal input field 

### LINKED ISSUE #667: Change Color feature is missing in the new design of Add & Update Goal

Tushar-4781:
Resolves #667


tijlleenders (changes requested):
There's an issue with the font color.
Can you fix that?

Also I expect the input field to change, not only the tags... but maybe that should be a separate PR.

tijlleenders:
It's very annoying I can't include screenshots from mobile. The upload fails every other day...

tijlleenders (approved):

tijlleenders:
Nice work!

The contrast is not great yet - but that will be fixed once we switch to the new colors I hope.

PR #675, opened 2022-09-30
Update: goal suggestions popup UI

### LINKED ISSUE #666: List the Goals from archive and public separately as per the new design

Tushar-4781:
Resolves #666


tijlleenders:
@Tushar-4781 This is great!

Can we make the archive/suggestions be part of the same list as the goals.
I mean can we make them scroll together instead of two separately scrollable sections?

image

tijlleenders:
After pressing + to add from archive/public - the archive/public appears.
But when you then press + again I expect the menu to appear. Instead the archive/public disappears.

tijlleenders (approved):
Looks good!
We'll phase out the popup once we decide how to differentiate the archive/public from the existing goals visually.

PR #683, opened 2022-10-05
Add script for retrieving latest wasm build.

egithinji:
This script is meant to be run from the root directory of a local clone of the `ZinZen` repo.

It checks the `ZinZen-Scheduler` repo to see if there is a new version of the wasm build. If there is, it retrieves it and creates a PR on the `ZinZen` repo to add the new build (i.e. update the `pkg` directory).

Before running it, the environment needs to have the following environmental variable set:
`export GITHUB_TOKEN={Your github token}`


egithinji:
Oh a couple more prerequisites I forgot to mention:
- Needs `Github CLI` to be installed
- Your repo should be using ssh authentication not https (otherwise it'll pause asking for username)

Tushar-4781:
> @Tushar-4781 Is this PR causing the Cypress tests to fail? It's only adding a bash script so I find that hard to believe.

@tijlleenders lemme check

PR #688, opened 2022-10-06
Update: dashed border input while editing the goal

### LINKED ISSUE #677: Highlight edit in place
Make edit a goal in place look like described in #671 (dots or stripes) whichever you prefer.

The goal sgould keep its original color - just the border is added.

Tushar-4781:
Resolves #677


tijlleenders (approved):
Big improvement in UX compared to jumping to a different screen!

We'll need to use it for a while to think about poasible design detail improvements (color, transition, hints, etc).

PR #707, opened 2022-10-13
Fix: Not detecting @ in start/by 12/3 @15

### LINKED ISSUE #706: Handle start and deadline @time
@Tushar-4781 I'm confused.

start and deadline are datetime fields right?
So we could type 'start 22/11 @13'.

If yes, this doesn't seem to be implemented yet so this issue is for adding that to the parser.

Tushar-4781:
Resolves #706

### DIFF
--- a/src/helpers/GoalTimingHandler.ts
+++ b/src/helpers/GoalTimingHandler.ts
@@ -42,7 +42,8 @@ function handleStart(lowercaseInput:string) {
       extractor: function extractDetail(_text : string) {
         const text = _text.split(" ").slice(0, 2).join(" ");
         const [e_date, e_month] = text.split(" ")[1].split("/");
-        return new Date(new Date().getFullYear(), parseInt(e_month) - 1, parseInt(e_date), text.includes("@") ? parseInt(text.split("@")[1]) : 0);
+        console.log(_text)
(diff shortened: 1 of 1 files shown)


tijlleenders:
Works for due but not for start.

Reproduce: 
start clean, edit exercise add @4 after start, doesn't work. 
Add @4 after due and only due works, see screenshot result:

tijlleenders:
Interestingly, removing the after tag solves the start issue.

Tushar-4781:
> Interestingly, removing the after tag solves the start issue.

@tijlleenders yes after is something you are mentioning explicitly just like inline CSS lol overrides the @

tijlleenders:
Ha, ok it was unexplained behavior :)
We'll have to think about what to do with impossible/weird combinations at some point... but not now.

Does the 'fix' still need to be merged?

Tushar-4781:
> Ha, ok it was unexplained behavior :) We'll have to think about what to do with impossible/weird combinations at some point... but not now.
> 
> Does the 'fix' still need to be merged?

its your call

tijlleenders:
Well if the changes don't actually fix anything - then no, I'll close the PR and issue. 

I've made #708 to think out if this is the right approach for the goalparser implementation.

PR #716, opened 2022-10-22
Fix: My Time page not rendering

### LINKED ISSUE #684: Fix MyTime default page
The default goals don't show up.
Not sure if this is since the new scheduler or if it was broken before.
signal-2022-10-06-061044

Tushar-4781:
Resolves #684


tijlleenders (changes requested):
There are schema issues with the scheduler that need to be solved. 

This issue is about getting the scheduler integrated with the MyTime based upon real Goals - not about creating dummy tasks for the MyTime page.

tijlleenders (approved):
Great to have this working!

PR #721, opened 2022-10-29
Update: sharing subgoal with parent goal name instead of root

### LINKED ISSUE #718: Make anonymously shared goal parentTitle dynamic
Right now the parentTitle of anonymously shared goal is always "root". This should be dynamic based upon the user's hierarchy.

So if I have a goal 'Learn Rust' and I anonoymously share a subgoal 'Read the Rust book' + URL then the parentTitle of the shared goal should be 'Learn Rust' - not 'root'.

Tushar-4781:
Resolves #718


tijlleenders (approved):
This is great!

Doesn't seem like much of a change - but this opens up a lot of possibilities already. :)

PR #724, opened 2022-10-30
Fix: Only non-null values in share goal object

### LINKED ISSUE #719: Remove any optional fields when sharing anonymously
Right now an anonymously shared goal is shared with all fields, even when they are optional and null: 
```
"body": "{"method":"shareGoal","parentTitle":"root","goal":{"title":"Check the Rust examples","duration":null,"repeat":null,"start":null,"due":null,"startTime":null,"endTime":null,"createdAt":"2022-09-09T22:00:00.000Z","goalColor":"#733D3D","language":"English","link":"https://doc.rust-lang.org/stable/rust-by-example/"}}"
```

Please don't share any optional null fields as this uses bandwith, compute and storage for nothing..

Tushar-4781:
Resolves #719

### DIFF
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -1,4 +1,6 @@
 /* eslint-disable no-alert */
+// @ts-nocheck
+
 import { db } from "@models";
 import { getJustDate } from "@src/utils";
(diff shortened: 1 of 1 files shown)


tijlleenders (approved):
Yup!

PR #728, opened 2022-10-30
Patch: Setting cursor to beginning for update goal

### LINKED ISSUE #720: Don't show tags in edit input
Problem:  
The edit input gets really crowded with a few tags. Cursor is at the end and you have to scroll through all tags to get to the title.

Solution:  
Don't show tags in edit input. This will keep editing nice and clean - restricted to the title.

New problem introduced:
How to edit the tags?
Solution:
For now if you want to edit the tag you remove the tag by clicking on it - then type again. It's not great but it is already implemented. In the future we might give a popup for editing or removing a single tag. So if I click the due date(time) I get a popup that allows me to edit the due date(time) - and only the due date(time) - or remove it.  
But for now just remove and re-add is a quick workaround to get some experience with it.

Tushar-4781:
Resolves #720


tijlleenders (approved):
Perfect!

PR #730, opened 2022-10-30

tijlleenders:
This PR was automatically generated by the `update_artifacts.sh` script.


tijlleenders:
GitHub finally is making attempts to provide fine-grained token permissions!
They still don't support the GraphQL API so had to do the last step manually.
Don't like using an admin token...

egithinji (approved):

PR #731, opened 2022-10-30
Update/update artifacts script

tijlleenders:
Few additions for our future selves :)

### DIFF
--- a/update_artifacts.sh
+++ b/update_artifacts.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
-echo Updating artifacts
-echo Checking if jq is installed...
+
+echo "Check prerequisites"
+echo "1. Github CLI installed"
+echo "2. Repo should be using ssh authentication - not https (otherwise it'll pause asking for username)"
+echo "3. jq installed"
+echo
+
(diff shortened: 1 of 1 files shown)


egithinji:
These are nice additions. I wonder if the wording here:
```
echo "Check prerequisites"
echo "1. Github CLI installed"
echo "2. Repo should be using ssh authentication - not https (otherwise it'll pause asking for username)"
echo "3. jq installed"
```
can be adjusted to "Please ensure the following:", or "This script requires:"
The "Check prerequisites" might give the impression that the script is checking if the prerequisites are installed, and then the subsequent lines are outputting the results of the checks.

update_artifacts.sh
```diff
@@ -1,12 +1,30 @@
 #!/bin/bash
-echo Updating artifacts
-echo Checking if jq is installed...
+
+echo "Check prerequisites"
```
tijlleenders, line 3:
```suggestion
Please ensure the following:
```

tijlleenders, line 3:
```suggestion
Please ensure the following:
```

PR #734, opened 2022-10-31
On deleting a goal remove it from archive as well && if archived remove it from parent sublist  

### LINKED ISSUE #725: Deleting goal should also remove it from archive
Deletion is permanent - you don't want it to show up anywhere anymore, not even in archive.

### LINKED ISSUE #732: Fix: Remove Goal from its parent sublist as well when archived
Currently after archiving the goal is not removed from the parent Sublist so in case the parent has only archived goal(s) the page is mounted and reverted back as we can't show archived goals. we need to prevent the mounting by removing Goal from its parent sublist as well when archived

Tushar-4781:
Resolves #725 
Resolves #732

### DIFF
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -24,14 +24,6 @@ export const addGoal = async (goalDetails: GoalItem) => {
   return newGoalId;
 };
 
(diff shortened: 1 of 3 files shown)


tijlleenders (approved):
This resolves the issues.

Did find a new one : #735

PR #739, opened 2022-11-03
Patch: goalItem interface variable naming & @ bug not appearing in input text 

### LINKED ISSUE #737: '@' not appearing while editing the goal

### LINKED ISSUE #738: Change the variable naming convention for datetime goalItem schema
startTime and endTime variable naming convention doesn't make sense its creating confusion in text used in tags. its better to replace the variable names similar to the tags

Tushar-4781:
Resolves #738 
Resolves #737


tijlleenders (approved):
Squashed!

PR #741, opened 2022-11-03
Update: Sync. of scheduler o/p with MyTime Page

### LINKED ISSUE #740: Fix the UI bugs on MyTime page
Scheduler's new build is released, the connection with the scheduler is working fine but the integration of scheduler output is pending which has introduced some bugs that can be fixed post synchronization of UI and o/p of scheduler

Tushar-4781:
Resolves #740


tijlleenders (approved):
Awesome!

PR #751, opened 2022-11-04
Update: Make Landing Page Mobile View as default for all screens

### LINKED ISSUE #694: Constrain screen size on bigger screens
We are designing for mobile first and don't have time to make a good responsive UI for bigger screens.  
Probably a bigger screen would require a significant redesign effort - to exploit the added screen real estate.

**Problem:** When opening on large screens the UI is suboptimal. Screen items are too far apart and not appropriately sized.

**Solution (for now):** We don't have time to fix this the right way.  
Even if it seems simple - it takes time and focus away from the MVP.
**Set a maximum screen size so the layout is always like on mobile, even on bigger screens.**

Below shows the problem when landing on the web app the first time:
image

Tushar-4781:


tijlleenders (approved):
Nice cleanup!

tijlleenders:
Fixes first of the onboarding screens issues in #694

PR #752, opened 2022-11-04
Patch: Keeping Mobile View UI as default view for all devices

### LINKED ISSUE #694: Constrain screen size on bigger screens

### LINKED ISSUE #709: Remove dashboard + switch Time/Goals with Logo
Problem: 
Navigation via dashboard is slow and not accessible on all pages (+ om right, logo in middle and back button left).

Solution:
Use the ZZ logo as a switcher between My Time and MyGoals.

Stretch goal if not too complicated: Swiping left/right also switches between the MyTime / MyGoals screens.

### LINKED ISSUE #753: Fix the overlapping of theme toggler and home button in navbar
image

Tushar-4781:
Resolves #694 
Resolves #709 
Resolves #753


tijlleenders:
Wow, that's some heavy editing!

Can you also remove the arrow and theme toggle on this page?
image

And here?:
image

My time page still too large:
image

Same for goals:
image

tijlleenders:
Corrected screenshots - the arrow and toggle can go completeley if we integrate #709 in this PR.

tijlleenders (approved):
Nice, I like it you kept the back arrow on the My Goals page!

PR #754, opened 2022-11-07
Add new icons

mallikarai05:
Added the icons for the My Goals page in .svg format.


tijlleenders (commented):
Feels like we're moving towards a style!

tijlleenders (approved):

PR #763, opened 2022-11-11
remove unused dependencies

tijlleenders:


Tushar-4781 (approved):
for now it looks good as everythin is working. will add if we find something missing

PR #790, opened 2022-11-21
shift the add options upwards #768

### LINKED ISSUE #768: Make batch add easier
Adding lots of items to a list one after the other (like when making a shopping list for the first time) has a little friction:
- after pressing + you need to move your finger down to + new for every item. This requires hand-eye coordination and takes focus away from the item to add.

Solution:
Move the dropdown up so that add new can be performed by two single taps on the + icon without moving the tap position:

mallikarai05:
resolved #768


mallikarai05:
@tijlleenders resolved the merge error

tijlleenders:
Nice.

Can you fix the alignment?

tijlleenders (approved):

Issue #779, opened 2022-11-20
Implement 2022-11 theme


mallikarai05:

mallikarai05:
![image](https://user-images.githubusercontent.com/73054648/202897577-f26bbef4-7720-4ee3-8e21-acd62a77179c.png)
I think we can have these as the palette for the dark and light themes for now. @tijlleenders @Tushar-4781 need your opinion for finalizing these.

Tushar-4781:
discussion on the enhancement of the themes can be discussed later but for now, it looks good I m only going to take the color palette and the rest other changes and even further modifications in UI will be made as soon as we are done with the current backlog of high prior tasks.. @tijlleenders what's your take?

PR #792, opened 2022-11-22
Theme Update: My Goals

Tushar-4781:
Part of #779


tijlleenders (approved):
Great to have this in so fast :)

PR #827, opened 2022-12-03
Fix: Navigation via goal tile

### LINKED ISSUE #793: Inrease 'tap' area for card handle
The currently 'dead' tap area between the goal title and the card handle (semicircle) should be added to the card handle area so that it's easier to uncollapse the card by pressing on/near the semicircle.

### LINKED ISSUE #819: Increase card tap area to whole card - not just title

Tushar-4781:
Resolves #793 
Resolves #819


tijlleenders (approved):
Nicely done!

PR #832, opened 2022-12-03
Feature: vibrate on selecting a choice on starter pages

### LINKED ISSUE #629: Add short vibration on initial choices
When selecting language, theme, intro - add a short vibration with navigator.vibrate

I believe this makes the experience more interactive and memorable.

Let's try it and see if it feels right.
Maybe we'll want to add haptic feedback in other places too.

Tushar-4781:
Resolves #629


tijlleenders (approved):

tijlleenders:
Doesn't work on DuckDuckGo but most users won't be using that anyway...

PR #834, opened 2022-12-04
Update: Raising Toast in hint option if no goals are there 

### LINKED ISSUE #799: Different bahavior for Public goals
The add 'Public' should be renamed 'Hint'.

When pressed a toast should be shown 'Looking for hints...'

The menu drawer from the bottom currently showing 'Sorry, no public goals' should only be shown if there are actually goals returned from the API call.

If no goals are found we don't have to show the drawer; we can toast 'Aww... We'll keep looking!'. 

We should store the fact that the call was unsuccesful and in the future, like a day later (and every day after that) - if the user navigates to the goal the API can be called again without user having to press + 'Hints' to check if any have been added - zz will automatically check. 

Actually the menu option for asking hints won't even be visible in that case since they are already asked and 'pending'.

All tags of the parentGoal should be added to request - **except the installId.** 
It's not necessary to tell 'the world' who is asking for a hint - it is sufficient to know someone out there wants help.

Tushar-4781:
Resolves #799

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -14,3 +14,14 @@
   background: linear-gradient(180deg, #121212 49.07%, #291B5C 100%);
   color: white !important;
 }
+
+#toast, #toast-dark {
+  position: absolute;
+  width: 95vw;
+  max-width: 600px;
+  top: 100px;
+  background-color: rgba(255, 255, 255);
+}
+#toast-dark {
+  background-color: #5468ff;
+}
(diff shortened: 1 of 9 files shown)


tijlleenders (approved):

tijlleenders:
Would be nicer to not show the Hint button at all if there are no hints.

Also - user needs to OK that we're sending the parent goal to ZinZen anonymously (without even the installId).

Issue #846, opened 2022-12-10
Change colors for better contrast


mallikarai05:
We need to improve these colors so that they are distinct, and the user can differentiate between colors to assign them to different goals.

mallikarai05:
![image](https://user-images.githubusercontent.com/73054648/206843851-668d18b5-01bf-4bce-ae91-d48cb3c3b9f0.png)
The present colors.

mallikarai05:
![image](https://user-images.githubusercontent.com/73054648/206857257-c82222e2-6055-446e-a1fb-a7db22146213.png)
These are the updated colors and pass the contrast test. 
@tijlleenders @Tushar-4781 need your opinion on these

tijlleenders:
Can't judge like this.
Let's start using these and evaluate in a week or two.

mallikarai05:
Will add these to the code for now and make a PR then!

PR #850, opened 2022-12-10
add new colors for goals

mallikarai05:
Added the new colors for goals as discussed in #846

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -31,9 +31,8 @@ export const getDiffInHours = (date1 : Date, date2: Date) => {
 };
 
 export const colorPallete = [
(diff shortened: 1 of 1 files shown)


tijlleenders:
@mallikarai05 is this PR ready to review?
Please ask for a review if it is.

What happens with my existing goals that have other colors if I merge this?

mallikarai05:
I changed the list of the previous colors only, @Tushar-4781 can you please look into it and let me know if there is something else that has to be done? Also, I think the colors of the present goals would remain the same I think Tushar would have more ideas about this.

Tushar-4781 (approved):
LGTM

PR #868, opened 2022-12-29
Feature: Goal Collaboration

### LINKED ISSUE #746: Share (suggest) complete
When a goal is known to be shared with someone else, there is an additional avatar:
image

When completing the goal, it first goes through a popup showing the options:
image

There's two checkboxes, the first one is ticked by default.

### LINKED ISSUE #747: Share (suggest) edit

### LINKED ISSUE #748: Share (suggest) delete

### LINKED ISSUE #750: Share (suggest) subGoal

### LINKED ISSUE #864: Develop & Integrate sharing/collaboration invites accordion design in sidebar

Tushar-4781:
Resolves #746  
Resolves #747  
Resolves #748 
Resolves #750 
Resolves #864


tijlleenders:
Copying the link to share with the contact doesn't work in Chrome nor DuckDuckGo.

Before it only worked in Chrome - now in neither.

Can't we use https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share ?

It is way easier to let the device present you with the apps that you can use to share and autopasting upon selection... than manually going to your app of choice and manually pasting.

tijlleenders:
Can we make the API URLs setting vars instead of hardcoding them?
This makes it easier to read and change.

tijlleenders:
Am I correct in noticing that there is no inbox/outbox for messages for processing changes asynchronously?

We said this would be necessary to allow for a reactive UI and offline sharing/collaboration actions.

tijlleenders:
The collaboration is now a flag/attribute on a goal - implying there is a single goal without room for storing conflicting views on attributes - all differences are either accepted or discarded, resulting in a "semi-shared" state / single goal.

This is in my view a fundamental departure from the ZinZen 'parallel' worlds approach - and I believe this will lead us down a path of exponential complexity.

I propose to keep any goal that "belongs" to others as a separate goal. Just register they both are part of a collaboration.
Any differences between my view/version on/of the collaborated goal and your view/version can then be easily identified - and accepted or rejected - and we can even change our mind at any time.

Trying to make "one version of the truth" is a lie - as we'll always have differing opinions. By merging changes selectively we are just fooling ourselves the differences don't exist.

Tushar-4781:
> Copying the link to share with the contact doesn't work in Chrome nor DuckDuckGo.
> 
> Before it only worked in Chrome - now in neither.
> 
> Can't we use https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share ?
> 
> It is way easier to let the device present you with the apps that you can use to share and autopasting upon selection... than manually going to your app of choice and manually pasting.

you can't copy link in incognito mode for inviting a contact. but if u create contact in normal mode it will work i.e the link will be copied

Tushar-4781:
> Am I correct in noticing that there is no inbox/outbox for messages for processing changes asynchronously?
> 
> We said this would be necessary to allow for a reactive UI and offline sharing/collaboration actions.

Commits are on the way. the PR still in progress. sorry forgot to draft it

Tushar-4781:
> 

yes, will clean the code at the end of the feature.

tijlleenders (approved):

PR #875, opened 2023-01-02
695-completing-goal-twice-adds-it-to-archive-twice

fendermany:
Have a good day!

I solved the problem.

PR #876, opened 2023-01-02
Fix: Completing goal twice adds it to archive twice

fendermany:
Fix #875

### DIFF
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -101,10 +101,10 @@ export const removeGoal = async (goalId: string) => {
     goals.forEach(async (ele) => {
       if (parentGoal === "root") {
         console.log("root");
-        if (ele.parentGoalId === "root") await db.goalsCollection.delete(ele.id);
+        if (ele.parentGoalId === "root" && ele.status === 0) await db.goalsCollection.delete(ele.id);
       } else {
         const tmpParentGoal = (await getGoal(ele.parentGoalId)).title;
-        if (tmpParentGoal === parentGoal.title) {
+        if (tmpParentGoal === parentGoal.title && ele.status === 0) {
           await db.goalsCollection.delete(ele.id);
         }
       }
--- a/src/api/GoalsAPI/index.ts
+++ b/src/api/GoalsAPI/index.ts
@@ -5,6 +5,7 @@ import { db } from "@models";
 import { colorPallete, getJustDate } from "@src/utils";
 import { GoalItem } from "@src/models/GoalItem";
 import { v4 as uuidv4 } from "uuid";
+import Omit = Cypress.Omit;
 
 export const resetDatabase = () =>
   db.transaction("rw", db.goalsCollection, async () => {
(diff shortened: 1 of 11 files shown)


tijlleenders:
Looking a lot better @fendermany !
An explanation of the problem and fix would be nice in the PR description.
You can ignore the Cypress tests - we need to update those....

@Tushar-4781 is it sensible for Dmitry to continue working on this PR given the changes we discussed today?
If not, can you propose some other issue for him to work on?

Tushar-4781:
@fendermany have a look on #877

src/api/GoalsAPI/index.ts
```diff
@@ -124,11 +111,17 @@ export const updateGoal = async (id: string, changes: object) => {
 
 export const archiveGoal = async (goal: GoalItem) => {
   const updatedGoalStatus = { status: 1 };
+  const parentGoal = await getGoal(goal.parentGoalId);
+  
+  if (parentGoal && parentGoal.collaboration.status !== 'none') {
```
Tushar-4781, line 116: Apologies, there are some modifications in the schema of goalItem as our story changed a bit. but approving for now

but Use **collaboration.status === "accept"** as that's the case where we will be following the exception

src/components/GoalsComponents/GoalSublistPage/GoalSublistPage.tsx
```diff
@@ -33,7 +41,7 @@ export const GoalSublist = () => {
   const callResetHistory = useSetRecoilState(resetGoalsHistory);
   const setShowAddGoalOptions = useSetRecoilState(displayAddGoalOptions);
 
-  const [tapCount, setTapCount] = useState(defaultTap);
+  const [tapCount, setTapCount] = useState<{ open: string | number, click: number }>(defaultTap);
```
Tushar-4781, line 44: Please do not engage yourself in refactoring. this creates conflicts in coming PR(s). Refactoring is done only when there is no **feature** in progress.

src/components/GoalsComponents/GoalSublistPage/GoalSublistPage.tsx
```diff
@@ -51,25 +59,21 @@ export const GoalSublist = () => {
...
 
-  const archiveMyGoal = async (goal: GoalItem) => {
-    await archiveUserGoal(goal);
-    await getChildrenGoals(goalID).then((fetchedGoals) => setChildrenGoals(fetchedGoals));
-  };
   const removeChildrenGoal = async (goalId: string) => {
     if (parentGoal?.sublist) {
       // delete subgoals of this goal
-      removeChildrenGoals(goalId);
+      await removeChildrenGoals(goalId);
```
Tushar-4781, line 76: we don't need to wait for this. as we are deleting the root goal in the next line, the sublist will be unreachable anyhow. so please do not use await here here

Tushar-4781 (changes requested):
@tijlleenders can you please check if the objective is achieved from UI point of view. Post that i will review rest of the files.

tijlleenders:
Did not test for shared or collaborated goal.
Think @Tushar-4781 can check this better.

For 'normal' (sub)goal the new behavior is quite nice from UX perspective!

The archived goal disappears from the list when you add it and can only appear once in the list with the same name. If you archive twice with the same title & parent - the last archived overrides the archived version with identical title&parent.

PR #879, opened 2023-01-03
update: passing duration as string in Scheduler

### LINKED ISSUE #878: Goals should use duration of type string
The recent build of the scheduler expects duration field of goals to be string. This build was merged but front end is still passing duration of type number, resulting in an error:

Screenshot from 2023-01-03 09-15-44

Tushar-4781:
Resolves #878


egithinji (commented):
nice

egithinji (approved):

Issue #742, opened 2022-11-04
Ask and show a whole week schedule on UI


tijlleenders:
Right now UI shows today, which is copy pasted to the other days of the week

Wanted: show the whole week's schedule, from Monday to and including Sunday. If the current day is not Monday, cross out the previous titles of the days.

egithinji:
@tijlleenders , @Tushar-4781 This is actually working if the fix committed by @Tushar-4781 here is merged: https://github.com/tijlleenders/ZinZen/commit/13c32f6e7ff9de36d36792c0dd309d9840c92501

But I had to delete my local ZinZen IndexedDB and refresh. There's nothing wrong on the scheduler backend.
![Screenshot from 2023-01-10 13-04-25](https://user-images.githubusercontent.com/51313777/211522807-202539fa-9038-48e6-b29a-a26536c502bc.png)

Tushar-4781:
@egithinji what we meant is that scheduler should give the schedule of entire week i.e. all seven days with what is scheduled and what's not ( the impossible ). Currently it return the schedule of the current day. and if u the same is repeated in entire week in the UI. @tijlleenders can u confirm this?

egithinji:
Please check comment on the above draft pr I've just made.
Remaining thing is to fix how the front end is putting all the tasks on the first day.

tijlleenders:
@egithinji has given the issue back.
Only thing left to do is some formatting on MyTime page - the data is ready and in the right format.

tijlleenders:
@Tushar-4781 #955 did break the UI as there is a new option array within every schesuled task returned. This shows a max three options for rescheduling, if any, and assuming everything else remains equal.

I guess we can add fixing that to the scope of this issue instead of making a separate one?

PR #890, opened 2023-01-05
Add helper module that filters tasks according to start day.

### LINKED ISSUE #889: Create module that filters tasks per day

egithinji:
Closes #889 .

### DIFF
--- a/src/helpers/TaskFilter/TaskFilter.test.ts
+++ b/src/helpers/TaskFilter/TaskFilter.test.ts
@@ -1,0 +1,133 @@
+import { TaskFilter } from "./TaskFilter";
+
+const t0 = {
+  taskid: "0",
+  goalid: "1",
+  title: "dentist",
+  duration: 1,
+  start: "2023-01-05T10:00:00",
+  deadline: "2023-01-05T11:00:00",
+  goalColor: "blue",
+  parentGoalId: "1",
+};
+
+const t1 = {
+  taskid: "1",
+  goalid: "2",
+  title: "shopping",
+  duration: 1,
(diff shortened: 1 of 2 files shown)


Tushar-4781 (commented):
It may not be serving the purpose. But would you like to try why the MyTime page is showing only 2 tasks?  earlier they all were schedulable now we see only 2.

egithinji:
> It may not be serving the purpose. But would you like to try why the MyTime page is showing only 2 tasks? earlier they all were schedulable now we see only 2.

Sure. Is there already an issue for this problem?

egithinji:
> It may not be serving the purpose. But would you like to try why the MyTime page is showing only 2 tasks? earlier they all were schedulable now we see only 2.

Sure. Is there already an issue for this problem?

tijlleenders:
@egithinji If integrating this module to the frontend is not part of this PR can you make a separate issue for that?

Is this test part of the CI now?

Tushar-4781:
> > It may not be serving the purpose. But would you like to try why the MyTime page is showing only 2 tasks? earlier they all were schedulable now we see only 2.
> 
> Sure. Is there already an issue for this problem?

nope

tijlleenders:
> > > It may not be serving the purpose. But would you like to try why the MyTime page is showing only 2 tasks? earlier they all were schedulable now we see only 2.
> > 
> > 
> > Sure. Is there already an issue for this problem?
> 
> nope

I think we don't need a new issue for this - just solve it as part of #742

Tushar-4781 (approved):
let's if these module can help else we will remove it in next refractoring

PR #920, opened 2023-01-10
Fix/my time page

egithinji:
Related to #742 .
There were three issues, 2 solved with this pr, 1 not yet solved.
1) The 'repeat' value was not being included in the input being passed to scheduler
2) The syntax used for 'repeat' wasn't what scheduler expects (should be all lower case i.e. "daily" not "Daily")

Now the scheduler gives the expected output, but the front end puts them all on the first day. This is the next thing to be fixed.

### DIFF
--- a/src/components/MyTimeComponents/MyTimeline.scss
+++ b/src/components/MyTimeComponents/MyTimeline.scss
@@ -1,4 +1,6 @@
 .MTL-display-light, .MTL-display-dark {
+  border: 2px solid black;
+  margin-bottom: 2em;
   padding-top: 1em;
   .MTL-taskTitle {
     color: #343434;
--- a/src/components/MyTimeComponents/MyTimeline.tsx
+++ b/src/components/MyTimeComponents/MyTimeline.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-key */
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useNavigate } from "react-router-dom";
 
 import { TaskItem } from "@src/models/TaskItem";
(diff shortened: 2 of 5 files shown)


Tushar-4781 (commented):
ok so now its returning 45 goals, is this how all 7 days schedule is returned ?

egithinji:
Yes if you look at the input and output, you can verify each but I did a spot check and believe it's returning what's expected. For example "Water the plants indoors" has three tasks on three different days, and the goals that are daily have a task for each day.

Tushar-4781:
> 
we are not expecting everything in one array. every day schedule should be separate.
{
dayName: "Tuesday",
Schedule: [ ... ]
impossible: [ ... ]
}
in the above format. so array of seven object each object has schedule of that respective day. let's have a quick call if something is unclear

egithinji:
@Tushar-4781 I don't think this is the best way of approaching displaying the 7 day timeline.

- It makes things very rigid and locks us into this 7 day display. For example we'll be prevented from extending the timeline beyond 7 days simply to avoid the problem of distinguishing between "Tuesday" of this week and "Tuesday" of another week.
- Grouping tasks for a specific day is more of a presentation/display issue and should be done on the front end for maximum maintainability and flexibility, not in the scheduler, whose responsibility is implementing the scheduling algorithm
- Working with strings like "Tuesday" is error-prone and less flexible than working with the actual `Date()`s or numbers of days

I suggest this logic gets worked into the front end, perhaps making use of the `TaskFilter` module.

Tushar-4781:
> @Tushar-4781 I don't think this is the best way of approaching displaying the 7 day timeline.
> 
> * It makes things very rigid and locks us into this 7 day display. For example we'll be prevented from extending the timeline beyond 7 days simply to avoid the problem of distinguishing between "Tuesday" of this week and "Tuesday" of another week.
> * Grouping tasks for a specific day is more of a presentation/display issue and should be done on the front end for maximum maintainability and flexibility, not in the scheduler, whose responsibility is implementing the scheduling algorithm
> * Working with strings like "Tuesday" is error-prone and less flexible than working with the actual `Date()`s or numbers of days
> 
> I suggest this logic gets worked into the front end, perhaps making use of the `TaskFilter` module.

ok so here are some comments that might resolve your doubts:
1. Nope its not rigid, if you want to view the next week or may be not 7 days but more than i will query with a startDate and endDate accordingly. i will fetch the next week to show the desired days or week.

2 & 3 no i m not bounding you to work on that day name like "Tuesday", you can put date too. rust can't extract the day name from date string/object ? I agree the taskFilter can be used, i m putting it as our last option but i recommend that we get output from scheduler in a structured format display above so we are just good to go. adding helper in between is a dependency which is not good for coming features as it will increase one layer of computation between and input and output.

lemme know if its still not possible then we will move forward with taskFilter function u made.

egithinji:
Please see this latest commit where I've implemented filtering on the front end.

I've added a dropdown to demonstrate how one can easily change the number of days to display and the app reactively displays the correct number of days.

Some of the styling (like the color lines) will need to be moved to the MyTimeline component and I haven't done that yet, but this should be sufficient to demonstrate the advantage of this approach.

Tushar-4781:
i don't know how to respond lol, @egithinji i didn't ask a proof, I understood your function at first place. I asked you whether its  possible for you to make the desired output at backend itself. I already mentioned that if its not possible then definitely we are gonna be using the helper function you created, I think you misunderstood.

tijlleenders:
@egithinji So looks like you have this, nice work.

Can you remove the dropdown and restore the page formatting to 'normal' so we can get this merged?

egithinji:
Please see #937

Tushar-4781:
@egithinji can we close this ?

egithinji:
@Tushar-4781 Yes

PR #925, opened 2023-01-11
Removed: Invite modal popup UI from share 1-1

### LINKED ISSUE #924: Remove current flow of collaboration from MVP

Tushar-4781:
Resolves #924 

The integration of colab functions in myGoals and its components are not removed. As soon as we will get the design of the new flow we will migrate the code from there to the respective file.


tijlleenders (approved):
Cheers!!

PR #926, opened 2023-01-13
fixed cypress tests 

### LINKED ISSUE #880: Fix Cypress tests based on roadmap

elyessAR:
Closes #880 .
modified/fixed the old tests to be compatible with our onboarding flow, testing the basic functionalities ( add, edit, search, archive and delete goals). also added cypress open script so other developers can visualize/debug the running tests locally with the "npm run cy:open" command.


elyessAR:
@Tushar-4781 the reason I used {force: true} in some clicks is because sometimes the button/input field gets covered by another element during the test 
ss2

you can see that search input is being covered by the nav bar, which fails the test.

Tushar-4781 (approved):
Nicely Done

PR #934, opened 2023-01-13
Refractoring | Fixed Minor BUGs | Small Patches | Code Optimization

### LINKED ISSUE #350: Logging out of the browser (clearing localstorage) gives 404 error
After deleting the localstorage i.e. basically logging out, the home button doesn't navigate to the landing page as shown in the gif below.

chrome-capture-2022-4-19

### LINKED ISSUE #507: SubGoals should inherit all properties except duration+repetition
SubGoals shpuld inherit all properties except duration+repetition:
- color
- language
- ... future properties as start and deadline

Duration and repetition are the exceptions and  are not inherited. 

Duration is not inherited so that you can subdivide a time budget into smaller parts. If you have a Gym workout of an hour and you put in subGoals - not all the subGoals should be an hour too, right. 
Repetition is not inherited. If you repeat your Gym workout daily, the repetition only has to be on the goal - not the subGoals. The prepocessor of the scheduler will handle repeating the tree as a whole.

### LINKED ISSUE #700: Increase readability of edit
Right now if you have a few constraints and edit the title goes offscreen and it is hard to read.

Wanted : show existing tags in their shortened form so there is less text and it is easier to read:
- "xh" instead of xhours
- remove any start today (will be added implicitly)

### LINKED ISSUE #817: White screen when viewing tags of hint
When asking for a Hint some public goals (bottom of the list Dentist and Boodschappen) give a white screen when attempting to view the tags.

Other work fine.

Could be a schema compatibility issue?
In that case removing these goals from the online hint library would be sufficient.

### LINKED ISSUE #833: Ellipse breadcrumb title issue when too long

### LINKED ISSUE #840: Change hint behavior
Only search for hint if explicitly requested.

Yes, this means:
- showing a loader while waiting for the internet instead of auto-searching hints with loader when pressing + on a goal 
- showing a 'Awww... no hints. We'll keep looking!' message when you're offline or if there are no results
- not showing the hint if there are no results
- keep looking for hints in the background every 24h

Tushar-4781:
Resolves #350 
Resolves #507
Resolves #700
Resolves #817 
Resolves #833 
Resolves #840


tijlleenders (changes requested):
Nice work.

Since there are multiple issues mixed together in this PR I've added comments in the respective issues that still need attention in my opinion:
#507 
#700

tijlleenders (approved):
Kaboom!

PR #937, opened 2023-01-14
Fix bugs affecting scheduler output

egithinji:
I suggest merging just these two commits (from #920).

This will fix the bugs that were resulting in wrong output from the scheduler.

@Tushar-4781 can implement filtering of the tasks per day on the front end, including any effects on styling, since he's more familiar with how the components fit together.


Tushar-4781 (approved):
LGTM

PR #970, opened 2023-01-28
update: keep inbox on right

### LINKED ISSUE #967: Keep inbox on the right

Tushar-4781:
Resolves #967


tijlleenders (approved):
Thanks, that was annoying.

The + is temoved on purpose? Can't we just switch to MyGoal as soon as you open the + menu, luke on MyTime?

I can imagine something in the inbox might trigger you to add something else to MyGoals...

PR #973, opened 2023-01-28
fix: refresh goals once deletion is done

### LINKED ISSUE #953: ui not updated when deleting multiple goals in a row
ui_not_updated_when_deleting_multiple_goals_in_a_row

Tushar-4781:
Resolves #953


tijlleenders (approved):
Great!

PR #974, opened 2023-01-28
fix: change goal title color to black in suggestions

### LINKED ISSUE #942: Change colour of text while adding from public goals

Tushar-4781:
Resolves #942


tijlleenders (approved):
Yes, now I can remove my 🕶️ 
😅

PR #977, opened 2023-01-28
Update: goals sorted by createdAt parameter

### LINKED ISSUE #893: Newly added goals appear at top of list
When I add a new goal that goal should appear at the top of the list. 

It might be interesting to apply some sort of sort index - as a preparation for custom drag&drop sorting - but I'm unsure how that should be implemented technically. Probably not on the goals themselves - as it's not their concern. Maybe in the parent goal as that already stores the children... although I'm pretty sure that that should go at some point in favour of storing relationships separately to allow for multiple parents. Probably a good enough place for now?

Tushar-4781:
Resolves #893


tijlleenders (approved):
Supernice :)

PR #980, opened 2023-02-02
Feature: Collaboration with new DB design and better UX

### LINKED ISSUE #960: Create Collaboration feature on top of new share 1:1 feature
As we have a new feature to display shared goals in Inbox. we need to set up new flow where we can convert the goal into collaboration as per the new Figma designs

Tushar-4781:
Resolves #960


tijlleenders (approved):
Long awaited, finally here!
Best way to test this is to merge and use the dev version with my partner.

PR #988, opened 2023-02-03
Fix: color bands width distribution | Update: day starts and end on 0:00

### LINKED ISSUE #966: Show timeline preview from 0:00 to 0:00 in 1h blocks
image

Space should be proportional to allocated time. So 24 equal units for 24 hours.

If proportional doesn't come out right the first 6 hours (0-6) of the day could get a little less space.

Tushar-4781:
Resolves #966


tijlleenders (approved):
Great!

PR #990, opened 2023-02-04
set starter goals at reasonable hours

tijlleenders:


tijlleenders:
#985 did not actually fix this.

Tushar-4781 (approved):

PR #991, opened 2023-02-04
nuance for privacy - sharing CAN happen explicitly

tijlleenders:


Tushar-4781 (approved):
Every word matters hahaha

PR #993, opened 2023-02-05
Fix/translate starter goals

### LINKED ISSUE #986: Translate starter goals
Make function ```addStarterGoal``` language-aware.

FYI For Dutch "Me time" can be translated to "Tijd voor mijzelf"

tijlleenders:
Resolves #986


tijlleenders:
@Tushar-4781 
I imported translations into MyTime instead of abstracting it in addStarterGoal function since I suspect we'll be using it for other things on this page, like 'Today' and 'Tomorrow'?

Shouldn't the translations be at an even higher level?
It seems overkill to load the library in every component that requires translations...  
Example: MyTimeline has buttons for 'Forget', 'Reschedule' and 'Done'.

Tushar-4781 (approved):

PR #1000, opened 2023-02-07
Feature: show archived goals in sublist in strike-through style 

### LINKED ISSUE #999: Line-through on collaborative/shared subgoals
The goal **archived** property should be true but it should be visible in the list havings its title crossed out. This is only applicable for subgoals belonging to the share or collaborative type.

Tushar-4781:
Resolves #999


tijlleenders (approved):
Nice!

PR #1001, opened 2023-02-07
Update: Show drop down icon to close task card 

### LINKED ISSUE #994: Whole MyTime card should respond to tap and collapse again if tapped 2nd time
The tap are of cards on MyTime should be like so:

In first tap - buttons forget / reschedule / done should show.

If tapped again (not on buttons) card should collapse again.

Tushar-4781:
Resolves #994

### DIFF
--- a/src/components/MyTimeComponents/MyTimeline.scss
+++ b/src/components/MyTimeComponents/MyTimeline.scss
@@ -118,3 +118,9 @@
   font-weight: bolder;
   color: inherit;
 }
+
+.task-dropdown, .task-dropdown-dark {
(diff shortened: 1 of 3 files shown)


tijlleenders:
@Tushar-4781 I like the dropdown but still woild like the whole card to be tappable to expand.

Tushar-4781:
Oh ok lemme do that

tijlleenders:
@Tushar-4781 Sweet that's what I meant.
Can you also make it work for collapse?

tijlleenders (approved):
Great!

PR #1012, opened 2023-02-11
Restored: correct code of MyTime page UI

### LINKED ISSUE #1007: CSS issue?
In #1001 or #1006 a css bug is introduced.
There is no bug in dark mode, only light mode.
A re-install / fresh incognito tab does not solve it.

image

Tushar-4781:
Resolves #1007


tijlleenders (approved):
Ah..... thanks!

Issue #316, opened 2022-05-13
Make the goal parser multilanguage


tijlleenders:
@GithinjiHans suggested we should make the goal parser in #273 multilanguage.
Right now it detects words such as "daily" and that is obviously different in another language.

Postponed this work to this issue as the current priority is to focus on making the app useable for at least English users.

tijlleenders:
The order of words can also change in other languages. In English we put the goal first, then the duration and other things like repeat, but in Dutch we start with the repeat, then the duration, and finally the goal...

tijlleenders:
We don't have a goal parser anymore.

PR #1029, opened 2023-02-19
Update/translations for goal text placeholder

tijlleenders:
Some fixes for a better placeholder text in English.
Some people actually typed in and shared "Walk + 1h + daily" with the actual + sign in there...


Tushar-4781 (commented):
I think translated placeholder is not a good alternative for people for now. for example, if they saw the translated version of **daily** tag then users will type that and the parser won't work because it accepts tags only in English.

tijlleenders:
Agree!
This PR is blocked by #316

PR #1031, opened 2023-02-19
Fix/remove unused files

tijlleenders:
Check the commits, I've removed some stuff that I think isn't used.
The assets file went from  
```dist/assets/index.9c7ad2b6.css                                  377.70 KiB / gzip: 59.01 KiB```
to  
```dist/assets/index.5a007e0b.css                                  36.48 KiB / gzip: 10.89 KiB```

Ok, I did download the bootstrap CSS from the CDN - but more often than not that will come from cache, or so they say...?
The CDN version is 28KB ... so also smaller. Seems like there's a 13ms impact, excluding cache. That's not the world - but it was fun to do.

Don't know - I'm completely new to this :). Is this a good idea or not?
Looks like everthing is working fine despite my slashing... but please double check.

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -31,14 +30,12 @@
     "recoil": "^0.7.1",
     "regenerator-runtime": "^0.13.9",
     "sass": "^1.50.0",
-    "uuidv4": "^6.2.13",
```
Tushar-4781, line 34: why we are removing this uuidv4?

Tushar-4781 (changes requested):
I think removing bootstrap css or something caused changes in the language selection screen in onboarding

package.json
tijlleenders: If I'm not mistaken we don't use it. We include v4 from lib uuid, not lib uuidv4.

tijlleenders:
Yes the language selection screen changed, but it is still useable - or not?

In any case, I don't know how to fix that and don't have the time/ambition to dive into it. Do you?
Should that block this PR?

package.json
Tushar-4781: uuidv4 itself contains an inbuilt function for generating uuidv5 too and we r still using uuidv4 in creating goals, contacts, and all.

PR #1043, opened 2023-02-24
Update: New Duration Pattern (15-17h) added in goal parser

### LINKED ISSUE #1042: Add "10-15h" pattern for duration in goal parser
Since, Scheduler takes duration as string, there is a request to add this pattern in goal parser

Tushar-4781:
Resolves #1042


tijlleenders (approved):
Awesome!!!

PR #1052, opened 2023-02-25
Update: Allow selecting changes in collaboration

### LINKED ISSUE #1050: Avoid switching: Refresh Goals immediately once changes are added in colab type goals
It's been observed that we need to switch pages to get the notification symbol. 
Expected Output: as soon as changes are added and the promise is resolved refresh the page content.

### LINKED ISSUE #1051: Allow users to select changes in collaboration
In Collaboration goals, currently, every change has to be saved or ignored. But it's time we release the 2nd version where users can choose what changes they want and save those only.

Tushar-4781:
Resolves #1050 
Resolves #1051

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -62,7 +62,7 @@ const App = () => {
                   .then(() => console.log("goal added in inbox"))
                   .catch((err) => console.log("Failed to add in inbox", err));
               } else if (["shared", "collaboration", "collaborationInvite"].includes(ele.type)) {
-                handleIncomingChanges(ele);
+                handleIncomingChanges(ele).then(() => setLastAction("goalNewUpdates"));
               }
             });
           }
--- a/src/Interfaces/IDisplayChangesModal.ts
+++ b/src/Interfaces/IDisplayChangesModal.ts
@@ -1,6 +1,20 @@
 import { GoalItem } from "@src/models/GoalItem";
 import { typeOfChange } from "@src/models/InboxItem";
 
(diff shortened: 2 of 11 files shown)


tijlleenders:
On initial share the link that is copied doesn't include the relId (Chrome).

tijlleenders:
The flow of sharing is very much improved with the explanation popups!

tijlleenders:
Great to see progress on this.
Let's dive in!

A makes goal.
A tries to share goal (creates contact B) and get popup if you understand 1:1.

Problem 1: if you click on accept you taken back to drawer instead of going to name contact input.

A clicks on + to create contact and shares link.

Problem 2 : link not copied with relId. => need to try sharing again an copy and communicate link with relId to share again.

B opens correct link and accepts.
A shares goal for real now.
B refreshes.

Problem 3: No indication there is something new in the inbox - but if B goes to inbox B will see the goal.

B makes this a collaboration.

Problem 4 : Switch to MyGoal but collaborated goal is not visible.

A makes a change.

Problem 5 : A refreshes but doesn't see the change.

B makes a change.

A refreshes.

Problem 6: A gets collaboration invite - but doesn't see the change from B.

Updates now working as expected in botb directions - with selective (un)checking of types of changes!

Unrelated:
Notification is hard to click on. Possible solution : open when uncollapsing goal that has the notification.

tijlleenders:
Sorry, 5 and 6 are not clear.
I'll make a video.

tijlleenders:
Agree that changes on shared goals don't require notifications. Maybe in the future we can only notify when there is a new shared goal or when it is deleted.
If you do want to see the changes, you can always make it a collaboration - even a one-sided one!

tijlleenders (approved):

PR #1065, opened 2023-03-14

tijlleenders:
This PR was automatically generated by the update_artifacts.sh script.


tijlleenders:
@Tushar-4781 Please check https://github.com/tijlleenders/ZinZen-scheduler/blob/main/tests/jsons/demo-3-with-filler-and-budget/input.json for the new input schema.

Some fields are optional, as you can see in https://github.com/tijlleenders/ZinZen-scheduler/blob/main/src/goal.rs

Tushar-4781 (approved):
Finally !!

PR #1074, opened 2023-03-22
Feature: Backup & Restore

### LINKED ISSUE #321: Make a 'Download my data' button/option

### LINKED ISSUE #322: Make a restore backup from file function
Will probably need a UI element since user has to search for a file.

For now just a function that can be called from the console to restore the whole repo from a file on the user's device.

The existing repo (if any) will get replaced.

The file is not encrypted at the moment - there will be a separate issue for that.

### LINKED ISSUE #323: Encrypt/decrypt backup of Goals/Feelings/...
When making a backup, the data collections should first be encrypted.

When restoring from a backup, the data collections should first be decrypted.

Key to use: min 5 character long, any visible character (so password123! is a valid key ans so is "passw").

### LINKED ISSUE #903: Add local download to side menu
Add a backup button to the side menu that generates a (link/button to) download to local file system.
Use https://dexie.org/docs/ExportImport/dexie-export-import is probably the easiest way.

The side menu is going to get long... but we'll organize it later.

### LINKED ISSUE #904: import from json blob
Add a backup import button to the side menu that imports json blob from local file system.
Use https://dexie.org/docs/ExportImport/dexie-export-import is probably the easiest way.

The side menu is going to get long... but we'll organize it later.

Tushar-4781:
Resolves #321 
Resolves #322
Resolves #323 
Resolves #904 
Resolves #903


tijlleenders (commented):
Awesome!

tijlleenders (approved):

PR #1085, opened 2023-03-24
Fix: blur feature should not be applied on habit in goal config

Tushar-4781:


tijlleenders (approved):
Hurray!

Going to production later today!

PR #1104, opened 2023-03-28
Fix: One side colab not working

### LINKED ISSUE #1097: Sharing goal should not require accepting collaboration
1. A initiates relationship with B
2. B accepts relationship
3. A shares 'Shopping list'
4. B sees 'Shopping list' in Inbox and makes it a collaboration
5. A adds subgoal in 'Shopping list' called 'Test 1'
6. B does not see 'Test1' ever
7. A has to accept turning shared goal into a collaboration
8. A now adds subgoal in 'Shopping list' called 'Test 2'
9. B now sees the subGoal added in 8, but not the one added in 6

sharing_requires_accepting_collaboration

Tushar-4781:
Resolves #1097 

Cause: It took me a while to figure out the bug, but I finally picked out the architecture diagram we discussed. this bug is because the sender treats the receiver as a shared type while it changed to collaboration. so we need to re-map the type receiver side.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -20,18 +20,19 @@ import { NotFoundPage } from "@pages/NotFoundPage/NotFoundPage";
 import { FeedbackPage } from "@pages/FeedbackPage/FeedbackPage";
 import { ShowFeelingsPage } from "@pages/ShowFeelingsPage/ShowFeelingsPage";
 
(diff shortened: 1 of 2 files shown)


tijlleenders (approved):
Awesome, I can even get subgoals shared before making it a collaboration.

PR #1107, opened 2023-03-29
Feature: Forget task in timeline

### LINKED ISSUE #1106: Implement the Forget Functionality in tasks
As per the discussion, forgetting is nothing but just canceling the task. So theoretically on UI, we will show it as forgotten while technically we will count the respective task's hours as spent so that scheduler can schedule the remaining hours of that particular task.

Tushar-4781:
Resolves #1106


tijlleenders (approved):
Awesome, like the sound!

PR #1119, opened 2023-04-07
update: reduced action(s) icon size

Tushar-4781:


tijlleenders:
Can we replace 'Complete' with 'Done'?

This will work out better for spacing.

tijlleenders (approved):

PR #1162, opened 2023-05-02
Fix: check & update dark mode and theme values when app opens

### LINKED ISSUE #1161: New theme not showing on dev branch for existing users
**Describe the bug**
The new themes is only showing for 'fresh browsers' on zinzen.vercel.app  
For browsers having visited the previous version, the old version keeps appearing.

**To Reproduce**
Go to a browser or a phone that had the previous version.
Refresh.

**Expected behavior**
New theme should appear automatically.

**Screenshots**
Old browser (ignore unrelated error):  
image

New release after hard refresh (checked the source - new files are loaded):  
image

**Additional context**
We need to fix this now before we get users.  
We can't keep resetting everyone's browser cache - since it also stores their goals.

Tushar-4781:
Resolves #1161

### DIFF
--- a/src/store/DarkModeState.ts
+++ b/src/store/DarkModeState.ts
@@ -1,6 +1,13 @@
 import { atom } from "recoil";
 
+export const getDarkModeValue = () => {
+  const mode = localStorage.getItem("darkMode");
+  if (mode && (mode === "on" || mode === "off")) {
+    return mode;
+  }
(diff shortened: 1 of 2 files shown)


tijlleenders (approved):

tijlleenders:
@Tushar-4781 
Yup, it works on refresh now!

PR #1180, opened 2023-05-03
improve confirmation modal UX, closes #1176

### LINKED ISSUE #1176: Improve confirmation modal UX
It's very annoying when you want to do something that seems simpel - but then get a popup with some text and choices that are unrelated to what you were trying to do. Example : Press complete goal button => Popup with some text and 'Confirm' or 'Cancel' - What am I confirming - I first have to read the text when I don't feel like it.

Instead of a generic 'Confirm' say 'Complete goal' like so:  
image

None.

**Additional context**
A generic confirmation modal is used for complete/delete/...  
This confirmation modal should be customizable so that the confirm text can be adapted to the context.

alpeshl:
Resolves #1176 
update context specific text to confirm dialog to make action more clear

### DIFF
--- a/src/common/ConfirmationModal.tsx
+++ b/src/common/ConfirmationModal.tsx
@@ -19,15 +19,21 @@ const ConfirmationModal: React.FC<ConfirmationModalProps> = ({ action, handleCli
     let confirmButtonText = "Confirm";
     switch (actionName) {
       case "archive":
-        confirmButtonText = `${confirmButtonText} goal`;
+        confirmButtonText = "Complete goal";
+        break;
+      case "delete":
+        confirmButtonText = "Delete goal";
         break;
       case "shareAnonymously":
+        confirmButtonText = "Share goal";
+        break;
       case "shareWithOne":
+        confirmButtonText = "Share goal 1:1";
+        break;
       case "colabRequest":
-        confirmButtonText = `${confirmButtonText} sharing goal`;
+        confirmButtonText = "Collaborate on goal";
         break;
       default:
-        break;
     }
 
     return confirmButtonText;
--- a/src/common/ConfirmationModal.tsx
+++ b/src/common/ConfirmationModal.tsx
@@ -1,4 +1,5 @@
 import { confirmationHeaders } from "@src/constants/confirmationHeaders";
+import { getConfirmButtonText } from "@src/constants/myGoals";
 import { ConfirmationModalProps } from "@src/Interfaces/IPopupModals";
 import { darkModeState, showConfirmation } from "@src/store";
 import { themeState } from "@src/store/ThemeState";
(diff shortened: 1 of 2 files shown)


alpeshl:
@tijlleenders I've pushed new updates as suggested

tijlleenders:
Thanks, that's fast!

Two changes and the  I thinj it's good.

1. You're missing a case statement for "complete" (don't know if that is the action name - please check). 

2. The wording for the "archive" action needs changing.

alpeshl:
@tijlleenders Just noticed one UI thing.
Not sure if that's intentional but just want to understand why this confirmation modal gets landed as aligned on left-top part of screen. 

Shouldn't that be positioned on center of screen to make it more accessible?

tijlleenders:
> @tijlleenders Just noticed one UI thing. Not sure if that's intentional but just want to understand why this confirmation modal gets landed as aligned on left-top part of screen.
> 
> Shouldn't that be positioned on center of screen to make it more accessible?

@alpeshl Yup, you're right to observe that is wrong!  
However, even though it seems like a small fix, all the browser UI things add up to a lot of hours, and some can be quite tricky - so we made the choice to completely ignore browser UI/rendering and focus on mobile first and only - at least for now. This gives us a little more time to work out the mobile.  
Mobile has the challenge that you have less screen space, no mouse-over etc... so it forces us to get down to the essentials.

tijlleenders:
@alpeshl Your comment about 'archive' in the code being 'Complete' in the UI pushed me to make #1181

tijlleenders:
@Tushar-4781 I think this can be merged from user-perspective. Can you double-check @alpeshl 's code?

src/common/ConfirmationModal.tsx
```diff
@@ -15,6 +15,30 @@ const ConfirmationModal: React.FC<ConfirmationModalProps> = ({ action, handleCli
   const [displayModal, setDisplayModal] = useRecoilState(showConfirmation);
   // @ts-ignore
   const { header, note } = confirmationHeaders[actionCategory][actionName];
+  const getConfirmButtonText = () => {
```
Tushar-4781, line 18: It is a simple function that gives us a constant value based on certain conditions and it will not be touched or used that frequently. I think we can move this function to constants -> myGoals.ts and use it from there. The goal is to keep .tsx files clean as much as possible

Tushar-4781 (changes requested):

src/common/ConfirmationModal.tsx
tijlleenders: @Tushar-4781 It is not clear to me what change you are requesting. The changes by @alpeshl resolve the issue.  

I do agree that is would be cleaner to have the context of what the button should say injected into the function as an argument - instead of the function having to know about all the different ways in which it is being used 'on the outside'. Is that what you meant? Or is it just about moving the current code as is to a different place?

alpeshl: I think what @Tushar-4781 is suggesting is to keep logic separate from UI aspects. getting button copy based on action context is something can be moved outside of UI file.

alpeshl: moved function to myGoal.ts

alpeshl: moved function to `myGoal.ts`

src/constants/myGoals.ts
```diff
@@ -26,3 +26,26 @@ export const getHeadingOfTag = (tagName: string, warningMessage = false) => {
...
+      confirmButtonText = "Complete goal";
+      break;
+    case "delete":
+      confirmButtonText = "Delete goal";
+      break;
+    case "shareAnonymously":
+      confirmButtonText = "Share goal";
+      break;
+    case "shareWithOne":
+      confirmButtonText = "Share goal 1:1";
```
tijlleenders, line 43:
```suggestion
      confirmButtonText = "Choose contact";
```
@Tushar-4781 rightly pointed out this gives the wrong expectation - we're not sharing yet ; we first have to pick a contact to share with.

tijlleenders (approved):

PR #1185, opened 2023-05-04
#1184 fix: Increased tap area for toggle

### LINKED ISSUE #1184: Increase tap area for theme toggle
It's very annoying when I have to aim my finger at a very small area of the screen.  

Make the tap area for the theme toggle bigger, like so:  
image  
Making the font size a little bigger and increasing separation between options would also help.

None.

**Additional context**
There is a UI guideline from Google that says that anything clickable on mobile should be a certain size - and have a minimum (thick finger) separation from another clickable thing.

alshareef49:
Resolves #1184 

Before fix:
Screenshot from 2023-05-04 16-09-08

After fix:
Screenshot from 2023-05-04 16-09-08

### DIFF
--- a/src/common/Header/Header.tsx
+++ b/src/common/Header/Header.tsx
@@ -30,6 +30,11 @@ const HeaderBtn = ({ path, alt } : {path: string, alt: string}) => {
   const [openInbox, setOpenInbox] = useRecoilState(displayInbox);
   const [darkModeStatus, setDarkModeStatus] = useRecoilState(darkModeState);
 
+  const toggleDarkModeStatus = () => {
+    localStorage.setItem("darkMode", darkModeStatus ? "off" : "on");
+    setDarkModeStatus(!darkModeStatus);
+  };
+
   const items: MenuProps["items"] = [
     ...["Donate", "Feedback", "Change Theme"].map((ele, index) => ({
       label: ele,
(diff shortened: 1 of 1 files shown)


tijlleenders (changes requested):
@alshareef49 Thanks for this!

The tap area is a lot better now.  
The other part is to make tapping the area actually change the theme - like the toggle button does.  
When I click the menu just disappears.  

I've converted the PR to draft - so you can request a review whenever you're ready - no rush.

alshareef49:
@tijlleenders now by just tap theme will change

tijlleenders (dismissed):
Yup, works for me!
@Tushar-4781 how about you?

src/common/Header/Header.tsx
```diff
@@ -47,7 +47,13 @@ const HeaderBtn = ({ path, alt } : {path: string, alt: string}) => {
     })),
     {
       label: (
-        <div style={{ display: "flex", justifyContent: "space-between" }}>
+        <div style={{ display: "flex", justifyContent: "space-between", marginTop:12 }}  onClick={
+          () => {
```
Tushar-4781, line 51: @alshareef49 the same code is used by the switcher as well. please make the following changes to pass the code - review.
1. Change that div to a button since you are trying to make it clickable 
2. Since two elements use the same code, create a separate function and then call it in the reqd. places

Tushar-4781 (changes requested):

src/common/Header/Header.tsx
alshareef49: @Tushar-4781 thanks for your instruction,
I made some changes  to the code according to requirements.

Tushar-4781 (approved):
Nicely done

alshareef49:
It's my first contribution to open source thank you guys 😀

PR #1199, opened 2023-05-07
Update: Scheduler api uses array of days in on_days

Tushar-4781:


tijlleenders (approved):
Thanks, works like a charm!

PR #1200, opened 2023-05-07
fix: invite link doesn't have a relId

Tushar-4781:


tijlleenders (approved):
Thanks, nice to have this in!

PR #1215, opened 2023-05-21
Update: New Edit/Add goal UI | Switching react-bs to antd

Tushar-4781:


tijlleenders (approved):
Great step up from previous edit menu!
Can't wait to get feedback on this.

PR #1225, opened 2023-05-27
UI Patches

Tushar-4781:


tijlleenders (approved):
Nice, this makes it a lot more intuitive!

PR #1229, opened 2023-06-01
Patch: Change Theme supports native back button

### LINKED ISSUE #1228: + button disappears
**Describe the bug**
After merge #1225 the floating + button disappears when pressing native back in theme switcher menu.
It does not come back - even after reload.

**To Reproduce**
1. Go to theme switcher
2. Click on native back button
3. No more floating + button

**Expected behavior**
Update so that + button comes back.

Tushar-4781:
Resolves #1228


tijlleenders (approved):
Thanks!

PR #1231, opened 2023-06-01
created: tests for app header

Tushar-4781:


tijlleenders (approved):
Based on quick search I guess it's good to check in the lock file too.

PR #1243, opened 2023-06-08
Patch: navigations & Update Now approach

### LINKED ISSUE #1228: + button disappears

Tushar-4781:
Resolves #1228 

Other issues that the PR resolves
- Bottom Navbar Tests
- Update Now Approach modified
- Bottom Navbar glitch of multiple selection
- Search bar connection with native back btn


tijlleenders (approved):

tijlleenders:
@Tushar-4781 
Awesome!

After merging this the + button came back in the inbox so I tried adding there. Then when closing Inbox I kept going back to the MyTime page instead of MyGoals.  
After closing/opening the app a second timr thr + button came back normally. It is also still visible in the Inbox - don't know if that was intended.  

Good to be able to add Goals again!!! 🎉

Tushar-4781:
@tijlleenders i tried reproducing the first issue. i m not redirected to MyTime i m still on my goals. regarding plus button's presence in inbox. I got a feedback that if while watching shared goal you just want to add the same without having collaboration then you can do by staying in inbox and even after goal is created you are still in inbox so that you can surf or watch remaining goals. lemme know if you want to revert this.

tijlleenders:
@Tushar-4781 
It is great this way.  
I understand the use case to have it in thr inbox and totally agree, thanks!

PR #1261, opened 2023-06-26
UI Patches

### LINKED ISSUE #1159: Make inbox disappear if empty
I always feel the urge to check the inbox icon if it's there. We have a nice image of it being empty when it's empty - but it would be even better if it's not even there when there are no messages.

Make the inbox icon disappear when the inbox is empty.

The current alternative is to show a nice message if inbox is empty, jay! :)
This is what most apps do - but I feel we can do better.

Tushar-4781:
Fix: Share with 1:1 not opening contact list
Fix: Loader not appearing on correct positioning in Add contact & Share goal modal

### DIFF
--- a/src/common/Loader.tsx
+++ b/src/common/Loader.tsx
@@ -10,10 +10,7 @@ const Loader = () => {
     <Spin
       className="loader"
       style={{ color: darkModeStatus ? "#705BBC" : "#CD6E51" }}
-    >
-      <span className="visually-hidden">Loading...</span>
-    </Spin>
-
+    />
   );
 };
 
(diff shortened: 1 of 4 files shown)


tijlleenders (approved):
@Tushar-4781 
The inbox is gone now indeed. Haven't tested if it appears again when there is a collab message.

I do see the inbox mail icon and the empty inbox picture 'flash' on screen for a split second upon refresh.

Tushar-4781:
This PR now doesn't resolve the inbox issue

tijlleenders (approved):

PR #1262, opened 2023-07-01
New Update: show inbox only when there are updates

### LINKED ISSUE #1159: Make inbox disappear if empty

Tushar-4781:
Resolves #1159 
Additional Updates:
- Inbox can be opened & closed from any level
- Search is prohibited when inbox is open
- if inbox is opened while you are nth level of a goal and closed the inbox in any way you will be back to that nth level state of the goal from which you opened the inbox


tijlleenders:
@Tushar-4781 

Hope a new developer will still understand this code... :)

When I accept a goal as a collaboration I go back to root level - which is expected behavior (I'm not closing inbox). However, when I then refresh I expect to still be at root level. Instead it takes me to the empty inbox page.

Tushar-4781:
> @Tushar-4781
> 
> Hope a new developer will still understand this code... :)
> 
> When I accept a goal as a collaboration I go back to root level - which is expected behavior (I'm not closing inbox). However, when I then refresh I expect to still be at root level. Instead it takes me to the empty inbox page. If I refresh again I'm still on inbox and need to press back twice to exit.

yeah the PR is bit complicated, i couldn't break it down as there was lot of back and forth between states, native back btn's & browser history

tijlleenders:
@Tushar-4781 

Allright so now the Inbox page stays open until I actively quit.  
I agree this is better, especially when you have multiple messages/requests to process.

tijlleenders (approved):

PR #1281, opened 2023-07-18
fix: Back button in MyGoals should move up first

### LINKED ISSUE #1279: Back button in MyGoals should move up first
**Describe the bug**

**To Reproduce**
1. Go to a sub goal
2. Click on MyTime or Journal
3. Go back to MyGoals
4. Press native back button or parent breadcrumb
5. See that you're not navigating 'up' the breadcrumb

vinaybadgujar102:
resolves #1279 

Changed navigation method to use navigate() instead of windows.history

### DIFF
--- a/src/components/GoalsComponents/GoalSublist/GoalHistory.tsx
+++ b/src/components/GoalsComponents/GoalSublist/GoalHistory.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { Breadcrumb } from "antd";
 import { useRecoilValue } from "recoil";
+import { useNavigate } from "react-router-dom";
 
 import { homeIcon } from "@src/assets";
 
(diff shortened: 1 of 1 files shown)


Tushar-4781 (changes requested):

tijlleenders:
@vinaybadgujar102 Thanks for working on this!  
It's a quite annoying and confusing situation. 

One thing to keep in mind is that the UI should remain snappy.  
Refreshing or redirecting with navigate doesn't feel snappy.

@Tushar-4781 Thanks for the suggestions!

Tushar-4781:
Sorry @vinaybadgujar102 i have to close this, as i already know where the the joint is and this isn't a bug it just a behavior that has to be overridden by our need. Thanks for making an attempt

PR #1294, opened 2023-07-22
Patch: JS scheduler handle three new ui options [ "once", per day,per week] from config ui

Tushar-4781:


tijlleenders (approved):
Should the Cypress test be updated?

Issue #1190, opened 2023-05-05
Minimize friction confirmation popup


tijlleenders:
**Is your feature request related to a problem? Please describe.**
It's very annoying when you press a button to do something and you get a bunch of text to read before being able to do/get the 'simple thing' you just asked for.

**Describe the solution you'd like**
Confirmation modal should highlight the most logical choice, and place it on the right side (based on the fact that a larger portion of us are right-handed).
<img width="307" alt="image" src="https://user-images.githubusercontent.com/44887755/236583896-3e38d0ac-ae62-480f-ae51-a1933e83d5d4.png">

**Describe alternatives you've considered**
None.

**Additional context**
Here's how that looks for some other screen:  
<img width="197" alt="image" src="https://user-images.githubusercontent.com/44887755/236584017-1f5dfac8-45c8-46b7-8ed2-58b6ac412d83.png">  
So to make it consistent the highlight in light theme should be this style:  
<img width="82" alt="image" src="https://user-images.githubusercontent.com/44887755/236584070-7ee60d24-4365-4332-91aa-bc5a54a67c63.png">

<img width="338" alt="image" src="https://user-images.githubusercontent.com/44887755/236584105-1ef265c8-1f7b-41a5-8c3f-86d1049162c6.png">

Tushar-4781:
@tijlleenders this can be close now right ?

tijlleenders:
@Tushar-4781 
#1295 Covers 95% of what I mentioned.
I guess we can make the UI overall consistent when a designer takes a look at it.

PR #1295, opened 2023-07-22
modal buttons position switched

Shubham-Praj:
This resolves issue #1190 

Old Modal Screenshot 
old modal ss

New Modal Screenshot 
UpdatedSS

### DIFF
--- a/src/common/ConfirmationModal.tsx
+++ b/src/common/ConfirmationModal.tsx
@@ -64,8 +64,8 @@ const ConfirmationModal: React.FC<ConfirmationModalProps> = ({ action, handleCli
         </Checkbox>
       </div>
       <div style={{ display: "flex", justifyContent: "space-around" }}>
-        { getChoiceButton(t(getConfirmButtonText(actionName))) }
         { getChoiceButton(t("cancel")) }
+        { getChoiceButton(t(getConfirmButtonText(actionName))) }
       </div>
     </Modal>
   );


tijlleenders:
Thanks @Shubham-Praj !

@Tushar-4781 Should the styling of the confirm button be white (like create goal and language choice) or pinkish (like confirn is now) in light theme?
It's not consistent...

Tushar-4781:
@tijlleenders button that perform certain actions or submissions are always supposed to vary from the entire layout background. it helps user to distinguish without putting much focus on text. if we make it consistent with the rest of color then user needs to put a little focus on both buttons to identify which one is **confirm**.  Lemme know if you want to alter that

Tushar-4781 (approved):

tijlleenders:
@Tushar-4781 That's not what I mean.
I agree with this - that's what I asked for. I mean the way we highlight certain buttons is different _across screens_. In light mode sometimes we highlight with white - sometimes with darker color.

Tushar-4781:
> @Tushar-4781 That's not what I mean. I agree with this - that's what I asked for. I mean the way we highlight certain buttons is different _across screens_. In light mode sometimes we highlight with white - sometimes with darker color.

yes its completely upto theme background and other component's background too like pop-up modals, bottom nav etc.
we have to keep minimum colors for each theme.

PR #1325, opened 2023-07-31
add: remaining translations for 'de' and 'pt'

vinaybadgujar102:
Added translations for the keys which were present in en but not in other languages.


vinaybadgujar102:
@tijlleenders @Tushar-4781

tijlleenders:
That is great @vinaybadgujar102 !

PR #1330, opened 2023-08-02
tijlleenders#1278: Change order of sub-menu buttons

jmorofsky:
Reordered the buttons. Also added tiny chevron separators as requested in comments.

### DIFF
--- a/src/components/BottomNavbar/BottomNavbar.scss
+++ b/src/components/BottomNavbar/BottomNavbar.scss
@@ -40,20 +40,6 @@
         font-weight: 500;
     }
 }
-#chevron::after {
-    position: absolute;
-    top: 26px;
-    left: 195px;
-    content: "";
-    display: inline-block;
-    width: 0.4em;
-    height: 0.4em;
-    border-right: 0.2em solid rgba(205, 110, 81, 1);
-    border-top: 0.2em solid rgba(205, 110, 81, 1);
-    transform: rotate(45deg);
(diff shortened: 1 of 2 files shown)


tijlleenders:
That's great @jmorofsky , thanks!

Could you remove the chevron? It's not really what I meant. The idea was to have the shape of the button itself as a rounded > on the right edge... but now I realize this will only be visible for the selected button unless we change the design to show separation between non-selected buttons.

So it looks good to me from UI perspective if you prune the commit for the >.

@Tushar-4781 will look at the code when he has time.

jmorofsky:
No problem, I have removed the chevrons. You could also think of some change to the design of the button. Maybe a border between the buttons so you could see the > even when its not selected, if you wanted to make that idea work.

tijlleenders:
Thanks @jmorofsky , LGTM!

As for your comment - that's indeed something to think about - but will be a different issue!

tijlleenders:
@jmorofsky Thanks a lot for this.  
I have been able to try this out in the preview link for a while ... and this made me realize that the order is actually fine the way it is because it's handy for the schedule to be on the left (most of us are righties).  

We'll just need to redesign the buttons like you suggested ... with the twist that it wouldn't go from left to right, but from right to left, so: Schedule <= Goals <= Feelings.  

This does require a new design though. I'll close this PR - but your contribution was useful to test this and realize the next iteration!

PR #1344, opened 2023-08-04
New Update: precache all binary files when pwa installs

Tushar-4781:


Tushar-4781:
Will tackle this issue after some time

PR #1357, opened 2023-08-06
Reference ZinZen-scheduler repository in README.md

kobe-reygel:
Add references between the Readmes of the main ZinZen repository and the ZinZen-scheduler repository to increase ease of navigation for visitors.


tijlleenders (approved):
Thanks @kobe-reygel !

PR #1366, opened 2023-08-06
Show minutes as superscript #1141

### LINKED ISSUE #1141: Show minutes on calendar as superscript
Right now the calendar time is shown as a long blob of text which is hard to read.

Example: "0:00-08:00"

It would be better to:
- remove any leading 0's ; so 08:00 => 8:00
- move the minute portion to superscript position ; so 8:00 => 800 (where 00 is in superscript)

Shahzaib-Iqbal-1057:
Resolves #1141

Hey! I have changed the way the time format looks : 
-Right now, I have removed the leading zeroes and also put the minute values in superscript, but the minute values won't be looking as good I think as all minute values are zeroes currently. 
-So, I have provided some more commented code with it as well in which only the leading zeroes are removed and there is no superscript. You can take a look at it if you want to.

### DIFF
--- a/src/components/MyTimeComponents/MyTimeline.scss
+++ b/src/components/MyTimeComponents/MyTimeline.scss
@@ -36,7 +36,7 @@
     width: 80px;
   }
   .MTL-goalTiming > sup {
-    vertical-align: text-top;
+    vertical-align:20%;
     font-size: 0.6rem;
   }
 }
--- a/src/components/MyTimeComponents/MyTimeline.scss
+++ b/src/components/MyTimeComponents/MyTimeline.scss
@@ -35,6 +35,10 @@
     white-space: nowrap;
     width: 80px;
   }
+  .MTL-goalTiming > sup {
+    vertical-align: text-top;
+    font-size: 0.6rem;
+  }
 }
 
 .MTL-display > button {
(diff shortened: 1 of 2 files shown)


tijlleenders:
Thanks for the effort @Shahzaib-Iqbal-1057 !

Doesn't it look better already?

It would be great if there's a way to make it pixel perfect. @Tushar-4781 any ideas?

Shahzaib-Iqbal-1057:
Try and decrease the vertical-align property by 3-4%  :)

Tushar-4781:
> Try and decrease the vertical-align property by 3-4% :)

go ahead. You can do this

Shahzaib-Iqbal-1057:
Take a look now

tijlleenders:
Perfect @Shahzaib-Iqbal-1057 , **exactly right!**

Do you need help resolving the merge conflicts?  

Also, could you align the hours with title?  
The first digit of the Tasks with two-digit starting hours should align left with the Task title.  
The time of Tasks with single-digit starting hours should align too, but with the - separator of the two-digit starting hours.  
Hope that explanation is clear... 😊

tijlleenders:
@Shahzaib-Iqbal-1057 Are you resubmitting this in a new PR?
There was just the alignment issue left...

tijlleenders:
@Tushar-4781 Please pick this one up. It's a shame to let the work go unmerged like this.

PR #1385, opened 2023-08-08
fix: removed share in public option

FullStackSorcerer:


tijlleenders:
Thanks @FullStackSorcerer !

That does the trick on the UI.  
There's some more code abd assets that can be removed but we'll leave that for another time.

tijlleenders (approved):

PR #1391, opened 2023-08-08
Show 100% done if there are no tasks

### LINKED ISSUE #1374: Show 100% done if there are no tasks today.

Senatrius:
PR to close #1374

### **Changes**
- Add a ternary operator to show done percentage if there are any tasks that day and show 100 percent if there aren't

### **Comment**
I am wondering if instead of showing "100% done" if there are no tasks added that day it would be better to show "No tasks today" instead, or something like that.

### DIFF
--- a/src/components/MyTimeComponents/ColorBands.tsx
+++ b/src/components/MyTimeComponents/ColorBands.tsx
@@ -55,7 +55,7 @@ const ColorBands: React.FC<IColorBands> = ({ list, tasksStatus, day, active }) =
       {active && day === "Today" && (
         <>
           <div className="task-progress">
-            <p>{`${Math.floor((dayCompleted / dayTotal) * 100)}% done`}</p>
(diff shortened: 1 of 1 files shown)


tijlleenders:
@Senatrius 
> I am wondering if instead of showing "100% done" if there are no tasks added that day it would be better to show "No tasks today" instead, or something like that.  

That would indeed be better. Good suggestion!

Tushar-4781:
hi @Senatrius, glad to see that you resolved that bug in addition to this can you add some padding below to this progress bar in case there are no tasks.

image

tijlleenders (approved):
Thanks a bunch ;)

Senatrius:
Replaced the message, now it shows "no tasks today" if no tasks were assigned that day. As for the padding, I've added it to the task list. 

My reasoning is that adding the padding to the list makes sure that it only applies when the task list is expanded and it doesn't add any extra height to the div when the list is collapsed.

tijlleenders:
Totally agree, nice one!

PR #1399, opened 2023-08-09
Navigate to goal hierarchy root with Goals button

### LINKED ISSUE #1381: Navigate home with Goals button
It's very annoying when you have to tap back multiple times when you are deep in a hierarchy.

When I'm already on MyGoals page and I tap Goals bottom menu bitton, take me to the root of the hierarchy (like pressing home in breadcrumb).

Use single tap to navigate up one step and double tap to go home - but that is confusing and you can just as easily use the back button.

Senatrius:
PR to close #1381 

### **Changes**
- Import `{goalsHistory}`
- Add an `if else` statement to go back to goal hierarchy root

### **Comment**
In the else statement I'm checking if goalsHistory length is not 0, otherwise clicking the Goals button reloads the entire page. It feels a bit strange but I'm not too familiar with Recoil.js. If there's a way around it or a better way to do it, I don't see it at the moment. If there isn't then good :)

### DIFF
--- a/src/components/BottomNavbar/BottomNavbar.tsx
+++ b/src/components/BottomNavbar/BottomNavbar.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import { useRecoilState, useRecoilValue } from "recoil";
+import { goalsHistory } from "@src/store/GoalsState";
 import { useLocation, useNavigate } from "react-router-dom";
 
 import backIcon from "@assets/images/backIcon.svg";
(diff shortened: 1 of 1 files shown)


tijlleenders:
@Senatrius Thanks, 

we'll take a look when there is time. Please be patient :)

Tushar-4781 (approved):
Nicely done @Senatrius , thanks for adding this little feature.

PR #1408, opened 2023-08-19
New Update: New designs of goal actions and edit goal

### LINKED ISSUE #1327: Design & Develop the MyGoal new UI
As discussed, it is crucial to track the time spent on tasks. If the user forgets to mark tasks as Done/Forget/Reschedule before the day ends, it's essential to present these unfinished tasks on the following day. This will allow the user to review and make decisions, ensuring that the hours are appropriately carried forward to the scheduler, facilitating a seamless schedule.

### LINKED ISSUE #1375: Change navigation behavior and remove 'add' sub-icon
It's very annoying when you want to edit a parent Goal and you have to remember to click on the half circle on the side instead of the Goal itself (because that goes into the goal).

Consistent behavior when clicking on a goal:  
- First tap opens the sub-menu
- Second tap goes 'into' the Goal.
  If there are no subGoals then it shows an empty screen.  

If this is implemented the sub-menu icon for "+ Add" sub-Goal is no longer needed and can be removed.  
Adding a sub-Goal can be achieved by clicking on the parent Goal twice - then use the floating + button bottom-right.

### LINKED ISSUE #1376: Use Goal circle for 'quick actions'
It's very annoying when you have many buttons in a sub-menu when clicking on a Goal.

When clicking on the circle next to the Goal title, show a standard popup with two options that are very easily clickable in a fast way (so big touch area's):
- Done (this will archive Goal)
- Delete (on right side of screen)

Remove the sub-Icons Delete and Done from the sub0menu.

### LINKED ISSUE #1405: Develop the Edit Screen design
Ref - #1363

Tushar-4781:
Resolves #1376 
Resolves #1375 
Resolves #1327 
Resolves #1405

### DIFF
--- a/src/components/GoalsComponents/InputGoal.tsx
+++ b/src/components/GoalsComponents/InputGoal.tsx
@@ -1,130 +1,130 @@
-import { useTranslation } from "react-i18next";
-import React, { useEffect, useState } from "react";
-import { useRecoilState, useRecoilValue } from "recoil";
+// import { useTranslation } from "react-i18next";
+// import React, { useEffect, useState } from "react";
+// import { useRecoilState, useRecoilValue } from "recoil";
 
-import { getDateInText } from "@src/utils";
-import { darkModeState, openInbox } from "@src/store";
-import { TagsExtractor } from "@src/helpers/TagsExtractor";
-import { extractedTitle, inputGoalTags } from "@src/store/GoalsState";
-import ITagExtractor, { ITagIndices } from "@src/Interfaces/ITagExtractor";
-import { createGoalObjectFromTags, formatTagsToText } from "@src/helpers/GoalProcessor";
+// import { getDateInText } from "@src/utils";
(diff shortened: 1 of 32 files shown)


src/components/GoalsComponents/InputGoal.tsx
tijlleenders, line 1: Shouldn't we just remove the dead code? 
If we need it we can always look in git.

tijlleenders (changes requested):
Yeah this is looking good!

A few adjustments required before we merge:d
- subgoals can't be reached
- checkmark after repeats weekly and no capital letters
- complete popup doesn't disappear after choosing complete or delete
- Spacing between circle and edit is too tight; it's hard to press one without hitting the other on mobile
- this was not in the issue but can we move the 'Create goal' button to the right of the hours field? That way we don't have to scroll all the way down for a quick add goal.

tijlleenders (approved):

PR #1436, opened 2023-08-25
Fixes #1433 Default values added to hours of the day

### LINKED ISSUE #1433: Pre-fill hours of the day by default
It's very annoying when I have to fill out fields when they can be pre-filled to fit the most common usecases.

Put real values (not a placeholder) in the hours of the day fields: 9-18

**Additional context**

This was part of #1363 but since the logic of differentiating between placeholder, suggestion and user-defined values and relationships for autocalculation was complicated we left out some default values. For hours of the day we can implement it though since we can use real values directly - not placeholders - and it is not dependant on other input fields.

Ktn-mariam:
### **What it does?**

fixes #1433


tijlleenders (approved):
Thanks @Ktn-mariam !

PR #1442, opened 2023-08-26
Create goal input hours fix

### LINKED ISSUE #1360: Validate/improve duration input
It's very annoying when I try to enter a duration as a new user and nothing gets scheduled.

Whenever I enter an invalid duration, like "0,5" hours, the scheduler should automatically round it to whole hours, with a minimum of 1. This should apply to any input field where hours are set - so not only the goal duration but also the max /day and max /week.

Show a dialog saying only whole hours are aloud - but that's just annoying. A toast could be nice, but then also auto-correct.

Shubham-Praj:
Resolves #1360 

Hi @Tushar-4781 & @tijlleenders 
This issue has been fixed. 

If input in decimal, it will be rounded off to the nearest whole number, with minimum value 1.

@tijlleenders, for the case of "0,5",where you mentioned that Dutch people use **comma** instead of **period**, this condition is already been handled, since the input type is set as **number**, it wont allow them to insert comma, while entering in hours.

### DIFF
--- a/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.tsx
+++ b/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.tsx
@@ -39,7 +39,7 @@ const CustomInput: React.FC<ICustomInputProps> = ({ placeholder, value, handleCh
 );
 
 const roundOffHours = (hrsValue: string) => {
-  return String(Math.max(Math.round(Number(hrsValue)), 1));
+  return hrsValue === "" ? "" : String(Math.max(Math.round(Number(hrsValue)), 0));
 };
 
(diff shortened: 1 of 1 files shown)


tijlleenders:
Thabks @Shubham-Praj 

There's a glitch: if I enter 0 it automatically puts in 1 (which is ok I guess) - but then after that if I try to delete it to have no value again it keeps putting the 1 back so I'm unable to change a 1 to a 2.

Could you check this?

Shubham-Praj:
@tijlleenders sure checking this

tijlleenders:
Great @Shubham-Praj !

@Tushar-4781 Can you merge if you agree?

Tushar-4781 (approved):
Nicely Done

Shubham-Praj:
Thanks 😬

PR #1447, opened 2023-08-30
New Feature: Partner mode

### LINKED ISSUE #1158: Add partner mode for browsing the other's goals
1. The current way of inviting a partner is quite complicated and involves a lot of steps - some of them going backward (you start by sharing but then get taken back to initiate a relationship first - after which you then have to share again.)
2. There is no good overview of what has been shared with you by your partner - and no insight in what their day/world looks like.

Add an avatar to the top left (to the left of header title).
When clicking on this you switch between me-mode and partner-mode.
Initially all pages will be empty - but with a message that invites you to share the link with your partner - so that you can share goals and idea's. 
This adresses 1 (makes inviting then sharing a more natural progression if share 1:1 is hidden until your partner accepts the invite). It also adresses 2.

The current design is the only alternative we've come up with so far - because we said that different 'modes' would be very confusing. I don't think the current design has a way of easily showing a peek into the partner's day/goals.

**Additional context**
Partner mode will some day be extended to 'others' mode - where you can also peek at your friends/colleagues/public figures etc world.

### LINKED ISSUE #1448: Notification dot is not at its place

### LINKED ISSUE #1449: Unable to initiate Collaboration

### LINKED ISSUE #1450: Sync the new goal schema to show changes in Collaboration

Tushar-4781:
Resolves #1158 
Resolves #1448 
Resolves #1449 
Resolves #1450


tijlleenders (approved):
Awesome!

PR #1495, opened 2023-09-16
fix: correct random index generation

### LINKED ISSUE #1437: Goal created without a color
**Describe the bug**
Sometimes a goal gets created without a color.

**To Reproduce**
Create a goal over and over - if you're lucky you can reproduce it.  
Probably better to look at the code and see if there is an 'off by one' error in the random generation or something.

**Expected behavior**
Color of the parent goal is maintained.  
If there is no parent goal the color is randomly selected.

**Additional context**
image

image

vinaybadgujar102:
resolves #1437 @tijlleenders


tijlleenders (approved):

tijlleenders:
Cool, thanks @vinaybadgujar102 !

PR #1496, opened 2023-09-16
fix: logo and MyGoal aligment

### LINKED ISSUE #1467: Increase logo size and align 'My goals' with Goal titles
It's very nice when things are in the right proportion and aligned.

Increase size of logo to same size as a Goal with subGoals icon.  
Align 'MyGoals' with Goal titles.

Keep logo size the way it is. It doesn't "feel" right. It's allowed to be a bit more prominent when you start the app and have only a simple list. Once you start using ZZ for a while you'll probably only have parent Goals on your root list. Then the logo will look small in conparison when it should equal.

**Additional context**

vinaybadgujar102:
resolves #1467 @tijlleenders

### DIFF
--- a/src/components/Header/Header.scss
+++ b/src/components/Header/Header.scss
@@ -7,7 +7,7 @@
   max-width: 600px;
   box-shadow: 0px 4px 6px rgba(0, 0, 0, 0.25);
   > :first-child {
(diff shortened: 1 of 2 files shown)


tijlleenders (approved):
Great!

PR #1509, opened 2023-09-19
refactor: replace old icon with new

### LINKED ISSUE #1470: Replace webmanifest logo's with new logo
It's very nice to use a consistent logo. We have a new one with two Z's in it and a more generally accepted color scheme.

Replace the current icons in webmanifest with new logo files. There's tools to automatically generate a whole set.

Keep it the way it is. Just a pity... we have a betyer logo - why not use it?

**Additional context**
Logo files on request.

vinaybadgujar102:
resolves #1470


vinaybadgujar102:
Hey @tijlleenders, please review and suggest changes to make

vite.config.ts
```diff
@@ -37,7 +37,7 @@ export default defineConfig({
         name: "ZinZen: Deliver purpose",
         icons: [
           {
-            src: "pwa-192x192.png",
+            src: "pwa-192x192-new.png",
```
tijlleenders, line 40: @vinaybadgujar102 Why rename the src if we're not using the old file anymore?

public/mstile-150x150.png
tijlleenders, line 1: @vinaybadgujar102 Shouldn't this have dimensions 150x150 as stated in the filename?

public/pwa-512x512.png
tijlleenders, line 1: @vinaybadgujar102 Why is the new file not filling the 512x512 area?

tijlleenders (changes requested):
It all looks good except some questions about sizes.
I don't know if that's related to the generator/program you used or if you did things manually.

https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs seems like a pretty recently updated exhaustive list of (fav)icons... but if you just regenerate the ones mentioned in review we can merge this.

public/pwa-512x512.png
vinaybadgujar102: I replaced it with the file which fills the full area, but as you can see the logo is not fully visible. The borders are cropped in icon tile. Same was happening with our previous icon.
WhatsApp Image 2023-09-20 at 10 55 42 AM
I have pushed new commit with new files, you can check it out.

public/mstile-150x150.png
vinaybadgujar102: this file was auto-generated using pwa-generator. The previous icon also had same dimensions.

vite.config.ts
vinaybadgujar102: resolved

public/mstile-150x150.png
vinaybadgujar102: According to the article you shared, now there is no need to create separate files for Microsoft tile support.

> Microsoft Edge used to [support](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/hh772707(v=vs.85)) a special icon format to pin websites to the start menu. For recent versions of Windows, this is no longer required.

Same applies for safari-pinned-icon.

public/pwa-512x512.png
tijlleenders: Well if this is the effect in practice then you were right to make it smaller!

tijlleenders (approved):

PR #1510, opened 2023-09-21
Fixed calender position issue when switching in sub menu 

### LINKED ISSUE #1480: Remember position Calendar when switching in sub menu
It's very annoying when you switch to your goals to check something and then it gets reset to the default week view.

- Set Today as the default view
- Remember if Today or 'This week' is selected when navigating away via bottom menu 

Default view only. Everyone is different. Just remember the last selected view.

**Additional context**
None.

vinaybadgujar102:
resolves #1480


tijlleenders (changes requested):
I have no idea what 'dailyViewAtom' means when I read it. This will make things harder to read/understand. 

Please choose a name tham makes it obvious what it does.  

Also don't use a boolean but state the state. Otherwise I have no idea what it means if this is false.

tijlleenders (approved):

PR #1513, opened 2023-09-22
add: sounds caching

### LINKED ISSUE #1289: Can't complete goal with flight mode on
**Describe the bug**
Sometimes I can't conplete a goal with flight mode on. 

**To Reproduce**
It's hard to reproduce consistently. Some goals will conplete - some won't. Sometimes navigating up and down the goal stack solves the issue.

I have a video but too large to post on GH.

vinaybadgujar102:
resolves #1289


tijlleenders:
I confirm it works.

@Tushar-4781 could you have a look at the code? We should still be on a strategy that minimizes loads and guarantees offline functionality - while still revalidating cache from time to time.

tijlleenders (changes requested):
@vinaybadgujar102 
I'm a little confused as we seem to have two locations where we are defining pre-caching in vite.config.ts:
- the workbox element with globPatterns
- the injectManifest with globPatterns

It seems from your commits that the first one is not doing anything?
https://vite-pwa-org.netlify.app/guide/service-worker-precache.html seems to indicate this way of pre-caching works with generateWS - not injectManifest strategy
What if you remove the workbox one completely - since it only seems relevant for generateSW?

Maybe then also add the other assets (png/svg/jpg) to the injectManifest?
... I doubt anyone would immediately go to flightmode after onboarding, but since the pre-caching is not slowing the visible load of the page (right?) we might as well precache that and enable a faster experience from there on out (on- or offline).

... then maybe also check the expiration plugin settings in `src/service-worker/service-worker.ts` as we might have to up the maximum entries a bit to not run into trouble if we add a few more assets: 
```
ll ./dist/assets/ | wc -l
54
```

tijlleenders (approved):

PR #1525, opened 2023-10-01
Goal title empty bug fix

### LINKED ISSUE #1523: Empty Goals are created using Spacebar
**Describe the issue**
When creating a new goal, it is possible to create an empty-looking goal by simply pressing the spacebar as the goal's title. This results in a goal that appears blank.

**To Reproduce**

1. Go to Goals page using bottom navigation.
2. Click on '+' button in bottom right corner
3. Press spacebar
4. Empty goal is created

**Expected behavior**
Users should not be able to create a goal with only spaces as its title.

Shubham-Praj:
Resolves #1523 

Added addition check condition for title in existing isEmptyTitle function.
Also removed the unnecessary isEmptyTitle function call from other functions.

Screenshot of toast when title is empty 
image


tijlleenders (changes requested):
Nice @Shubham-Praj !

Can you also reset the goal text to "" so the placeholder appears after the toast?

Shubham-Praj:
Yes sure

Shubham-Praj:
Hey @tijlleenders placeholder part is fixed

tijlleenders (approved):
Awesome @Shubham-Praj !

PR #1530, opened 2023-10-03
Auto reduce max duration to 99 issue fix

### LINKED ISSUE #1443: Auto reduce max duration to 99
It's very annoying when you accidentally enter a three digit duration like 120 when you actually wanted to enter 20. It then shows like 20 on the edit screen when in fact it's 120.

Whenever a duration >99 is entered - autocorrect to 99.

Allow durations of >99 and show 3 digits. This has several problems. It isn't easy to work/think with so many hours. If we need a big duration we'd like to enter days, weeks or months. Also it's not a common use case - when there are alternatives: use a repeat + due date or break down the work jnto max 99h subGoals.

Shubham-Praj:
Resolves #1443 

All the underlined field in modal, will have this feature
image


tijlleenders (approved):
Excellent, thanks for the detailed PR comment @Shubham-Praj !

PR #1576, opened 2023-10-28
edit: Update goal circle icon representation

### LINKED ISSUE #1568: Make goal circle used for leaf-goal bigger
After differentiating between Goals and Budgets we need a visual cue to show the difference in the goal circle icon.  
This is complicated as the goal icon is quite small, and the way to show sub-goals is to have a bigger circle around it; which would make recognizing the budget icon harder.

Before making a special goal circle for a budget, switch the logic of signalling sub-goals. Instead of showing a bigger goal circle around the parent goal circle, start with a bigger circle by default for leaf-goals and add a smaller goal circle inside of if in case the goal gets sub-goals.  This seems more intuitive - as the parent 'contains' the sub-goals. It will also be easier to recognize a budget from a goal - while both could contain sub-goals/budgets.

None

**Additional context**
Leaf goal before requested change:  
image
This circle has to change to be as big as outer circle of goal with sub-goals (see below).

Goal that has sub-goals before requested change (ignore misalignment; it doesn't happen on mobile):  
image
This representation wouldn't change.

vinaybadgujar102:
resolves #1568

### DIFF
--- a/src/components/GoalsComponents/MyGoal.tsx
+++ b/src/components/GoalsComponents/MyGoal.tsx
@@ -144,6 +144,18 @@ const MyGoal: React.FC<MyGoalProps> = ({ goal, showActions, setShowActions }) =>
     }
   }, [location]);
 
+  const hasSubGoals = goal.sublist.length > 0;
+
+  const outerBackground = hasSubGoals
+    ? "transparent"
+    : `radial-gradient(50% 50% at 50% 50%, ${goal.goalColor}33 79.17%, ${goal.goalColor} 100%)`;
+
+  const innerBackground = hasSubGoals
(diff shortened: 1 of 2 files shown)


tijlleenders:
@vinaybadgujar102 the outer circle (with the shading) now seems to get smaller when there are subGoals. This is confusing.  

Can you make it so that the outer circle remains the same, but only a small circle without shading gets added its middle when it gets subGoals?

tijlleenders:
LGTM @Tushar-4781 ?

Tushar-4781 (approved):
Code looks good, btw actions text is barely visible, do we need it ?

tijlleenders:
It's not very readable, I agree. You could zoom in. Another option is to replace it with three dots which are easier to recognize without zooming in. Different issue though!

PR #1581, opened 2023-10-29
feat: confirm edit goal by pressing enter

### LINKED ISSUE #1077: Confirm new edit goal with enter
Rewritten based on discussion below:

We've gone back an forth on this but a consistent behavior is better:  
- A keyboard 'Enter' on mobile or desktop browser should have the same effect.
- The effect should be the same as the 'action button' - in this case the 'create goal/budget' button. It should not only close the keyboard on mobile.

Agree with @vinaybadgujar102 we can add a test for this in Playwright.

vinaybadgujar102:
resolves #1077


tijlleenders:
@vinaybadgujar102 Did you test on mobile?  I get "Next" on my keyboard and it does what it says : move the cursor to duration field.

Tushar-4781 (approved):

PR #1585, opened 2023-11-01
fix: drag event not working

### LINKED ISSUE #1583: Can't drag & drop twice in same list
**Describe the bug**
After I drag & drop one item in a list it stays 'stuck/highlighted' and I can't drag & drop anything else.  
The first drag & drop is also not persisted.

**To Reproduce**

1. Go to a list with multiple items
2. Drag and drop to change position of one item
3. See it stays highlighted
4. Try to drag & drop something else - you can't
5. Navigate away from list and cone back again - drag & drop from before is undone.

**Expected behavior**
Drag & drop can be done multiple times.  
Persisted at each time.

**Screenshots**
None.

**Additional context**
None.

vinaybadgujar102:
resolves #1583

### DIFF
--- a/src/components/GoalsComponents/GoalsList.tsx
+++ b/src/components/GoalsComponents/GoalsList.tsx
@@ -49,7 +49,8 @@ const GoalsList: React.FC<GoalsListProps> = ({ goals, showActions, setGoals, set
     Promise.all(posIndexPromises).catch((err) => console.log("error in sorting", err));
   };
   return goals.map((goal: GoalItem, index: number) => (
-    <React.Fragment key={goal.id}>
+    // eslint-disable-next-line react/no-array-index-key
+    <React.Fragment key={index}>
       {showUpdateGoal?.goalId === goal.id && <ConfigGoal action="Update" goal={goal} />}
       <DragAndDrop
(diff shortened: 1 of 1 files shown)


tijlleenders:
Seems to work - but why are we bypassing ESLint?

vinaybadgujar102:
@tijlleenders the issue was with goal-id being used as key value. When goal-id is used as a key, React assumes that the order of items is stable. 
so there were two choices, either to drop key prop or to use other value. I could only think of index as key and it was throwing the eslint error.

Tushar-4781 (changes requested):
hi @vinaybadgujar102  try this 
import { v4 as uuidv4 } from "uuid";
key={uuidv4()}

vinaybadgujar102:
@Tushar-4781 it's not working

Tushar-4781:
> @Tushar-4781 it's not working

hi @vinaybadgujar102 what's the issue you are facing

vinaybadgujar102:
@Tushar-4781If we use key={uuidv4()}, then drag and drop is not working at all. I think using uuidv4() as a key generates new values at every render.

Tushar-4781 (approved):
Let's deal with this later

PR #1619, opened 2023-11-23
send all budgets and goals with childrens to scheduler

### LINKED ISSUE #1594: Send all Budgets and Goals with duration - including all ancestors - to scheduler
It's very nice if the scheduler works.  
Currently not everything that is needed gets sent to the scheduler.

Send all Goals with a duration and all Budgets to the scheduler.
For both - include all of their ancestors, even if they are Goals without a duration.

Filtering out empty stuff. This could be done as an improvement later.

**Additional context**
The logic has changed. This is not fixing a bug - just changing the approach.

vinaybadgujar102:
resolves #1594 
@Tushar-4781 @tijlleenders not very confident with this pr as I am less familiar with scheduler logic, but tried.


tijlleenders:
@vinaybadgujar102 
Nice start!

This is a little tricky... When I make a root Goal without duration - but then a sub Goal with duration - I do expect both to be sent.
image

@Tushar-4781 What is a good technical trick to filter out any (sub-)trees that entirely have no duration?

tijlleenders (changes requested):
@vinaybadgujar102 
Nice start!

This is a little tricky... When I make a root Goal without duration - but then a sub Goal with duration - I do expect both to be sent.
image

@Tushar-4781 What is a good technical trick to filter out any (sub-)trees that entirely have no duration?

vinaybadgujar102:
@tijlleenders now check if all the scenarios are handled correctly or not

tijlleenders (dismissed):

Tushar-4781 (changes requested):
@vinaybadgujar102  Nice try!!. But let's send all the goals to the scheduler and let rust scheduler do the filtering and all this heavy lifting. :)

tijlleenders:
Allright - well that should simplify things in TS land.

Tushar-4781 (approved):

PR #1628, opened 2023-11-26
improve goal adding ux

### LINKED ISSUE #1618: Improve Goal adding UX
It's very nice when I don't have to move my finger to two places when quickly adding a Goal. The solution we have now is that the + button turns into a Goal circle with a hardly readable (also because of bad contrast) "Goal" text.

However, if you don't know that - it seems like you can only add a budget - because your finger is blocking the view of the Goal circle...

Expand the + button to the left. This will show you what you're clicking - without moving your finger - and allows for a bigger font size.

Expanding to top. This doesn't allow for bigger font size.

**Additional context**

### LINKED ISSUE #1644: Translate action button - at least for Dutch
It's very nice to have all of the interface in Dutch when the app is being featured in a Dutch movie (hitting Cinema in February 2024 so not a lot of time left...).

Translate the action button to at least Dutch.
Goal => Doel
Budget =>  Budget

None.

**Additional context**
image

vinaybadgujar102:
resolves #1618 
resolves #1644


tijlleenders:
@vinaybadgujar102 
The + button is not visible on DuckDuckGo (mobile). I don't think we should waste too much time on tweaking minor UI details for different browsers - but the basic elements should be visible...

Also, can you use a consistent font size (like the ones used for Goal text)?  
Having different font sizes mixed on screen makes it noisy and harder to read.

It would be nice if there is no gap between the 'Goal button' and the '+' button. That will also avoid the + button from 'jumping' to a different place after initial tap.

vinaybadgujar102:
@tijlleenders I have made changes as mentioned

tijlleenders:
Yeah nice @Tushar-4781 OK to merge?

src/components/BottomNavbar/BottomNavbar.tsx
```diff
@@ -120,8 +121,8 @@ const BottomNavbar = ({ title }: { title: string }) => {
             title={themeSelection ? "ArrowIcon" : "JournalIcon"}
           />
           {themeSelection ? <p>Next</p> : <p>{t("Journal")}</p>}
-          {isAddBtnVisible && <GlobalAddBtn add={title} />}
         </button>
+        {isAddBtnVisible && <GlobalAddBtn add={title} />}
```
Tushar-4781, line 125: hi @vinaybadgujar102  please put this back inside the journal button. It was done intentionally. as we have to keep the Add button right above to the journal icon. if you move this button out we will loose the alignment on mobile screens. for example this device
image

Tushar-4781 (changes requested):

vinaybadgujar102:
resolves #1644

Tushar-4781 (approved):

PR #1633, opened 2023-11-27
#1631 Issue: Replace 'Following' button with a switch

### LINKED ISSUE #1631: Replace 'Following' button with a switch
It's very annoying when a button is not clear. Right now the 'Following' button has the action color:
image
... but when I unfollow it has a near-identical appearance:  
image

Replace the "Following" button: 
image
... with a switch:
image
+ to make the switch make sense change the title of the modal to 'Following' instead of participants.

None.

**Additional context**
None.

siva9495:
Resolves #1631 
Modified title of modal Participants to Following.
Checked dark mode is OK.


tijlleenders:
@siva9495 This is great, thank you!

There is one little detail: Upon initial opening the toggle state does not sync with the database info:
image

After I toggle once, they are in sync. Is that something you could investigate?

siva9495:
@tijlleenders  I have made changes as mentioned

Tushar-4781 (changes requested):
hi @siva9495 i really appreciate your efforts here. Unfortunately, i would like to discard this solution as we are already using antd library. you can use the switch component from the same. https://ant.design/components/switch you can pick the code and modify it as per the description written in the issue

siva9495:
HI @Tushar-4781 
As mentioned above updated the code using the antd library for the toggle switch.

package.json
```diff
@@ -8,6 +8,7 @@
     ]
   },
   "dependencies": {
+    "@ant-design/icons": "^5.2.6",
```
Tushar-4781, line 11: @siva9495 i don't think its right time to add this dependency. you can use the existing svgs for this. if you open the actions popup my pressing on the circle you will find the two icons. use them instead :) 
image

Tushar-4781 (changes requested):

package.json
siva9495: @Tushar-4781 I have a doubt regarding whether the SVG needs to have the same color or be white. Please confirm.

Tushar-4781: its better to use the svgs like we are using in other components. i don't understand the sizing issue you mentioned. can you put some ss of code or output. thanks.

siva9495: @Tushar-4781 , sorted it. Removed the library and fixed it using SVGs for the icons. Can you please confirm?

Tushar-4781 (approved):

tijlleenders:
Thanks @siva9495 glad we got your contribution in!

PR #1642, opened 2023-12-01
simplify my time nav

### LINKED ISSUE #1623: Simplify navigation on My Time
It's very annoying when navigation is confusing. Right now the behavior of the back button on My Time is not intuitive. There are three screens that act as one. 

Consistent navigation between screens using only the bottom navigation and two levels on My Time (the main calendar screen and a focus screen).  
This requires merging the This week screen into the main My time screen.
This week:
image

Main My time screen:
image

The focus mode would then only be activated by clicking 'Focus' on a Task sub-menu.  
Pressing back from that screen will take you back to the main My time screen.

None.

**Additional context**
We moved the This week because it was too busy. We'll have to find a different solution to that as moving it to a separate screen is hard to use.

vinaybadgujar102:
resolves #1623


tijlleenders (changes requested):
Nice job @vinaybadgujar102 !

We're ...nearly... there. Just a minor glitch to fix:
- make a new goal with a duration
- go to My Time and 'Focus'
- press back twice
- go back to My Time
The Task is now completed ... even though I never manually completed the Task.

Could you look into this?

tijlleenders (dismissed):

Tushar-4781 (approved):

PR #1646, opened 2023-12-01
keep zinzen text logo visible

### LINKED ISSUE #1643: Keep ZinZen text logo visible while adding goals
It's very nice when we can see the name ZinZen on screen when adding goals - especially since we're going to be featured in a movie and that's the only way to communicate the app name to people watching the movie.

Keep the ZinZen text logo while adding goals.
As soon as goals nearly overlap with ZinZen text logo, push the logo downward a little.

Show ZinZen permanently on header. That's just annoying.

**Additional context**
image

vinaybadgujar102:
resolves #1643

### DIFF
--- a/src/pages/GoalsPage/MyGoals.tsx
+++ b/src/pages/GoalsPage/MyGoals.tsx
@@ -68,6 +68,8 @@ export const MyGoals = () => {
     }, 300);
   };
 
+  const zinZenLogoHeight = activeGoals.length > 0 ? "125px" : "350px";
+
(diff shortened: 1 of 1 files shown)


Tushar-4781 (commented):
Code looks ok @tijlleenders can you confirm if this is the behavior that you were expecting

tijlleenders (commented):
@Tushar-4781 Thanks, yes this is what was asked!

tijlleenders (approved):

PR #1651, opened 2023-12-02
goal add button positioning

### LINKED ISSUE #1627: Correct placement of add button in browser
It's very annoying when somebody opens ZinZen on browser for the first time - and doesn't see the add button because it's way outside of the main area.

Correctly place the + button on browser, just like on mobile

Leaving it the way it is - as we're designing for mobile only.  
This is however a quick-fix - and can help with the first contact that for some people will be on browser.

**Additional context**
We choose to support mobile-first and mobile-only - so that we can focus and get more done.
image

vinaybadgujar102:
resolves #1627


tijlleenders (approved):
LTGM @Tushar-4781 OK on code?

PR #1672, opened 2023-12-07
add flip animation to ZZ logo when switching partner mode

### LINKED ISSUE #1591: Flip animation on ZZ logo when switching to partner
It's very nice to feel/see something changes when switching to partner mode.

A quick flip or turn animation when switching to partner mode. 

None.

**Additional context**
Can probably be done with standard css animations.

vinaybadgujar102:
resolves #1591

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -70,7 +70,7 @@ const Header: React.FC<IHeader> = ({ title, debounceSearch }) => {
   useEffect(() => {
     const timer = isFlipping ? setTimeout(() => setIsFlipping(false), 500) : undefined;
     return () => clearTimeout(timer);
-  }, [isFlipping, setIsFlipping]);
+  }, [isFlipping]);
 
   const currentHour = new Date().getHours();
   const isNighttime = currentHour >= 18 || currentHour < 6;
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -1,5 +1,5 @@
 import { useTranslation } from "react-i18next";
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { useLocation, useNavigate } from "react-router-dom";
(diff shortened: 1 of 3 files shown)


tijlleenders:
Looks awesome @vinaybadgujar102 !

src/components/Header/Header.tsx
```diff
@@ -65,6 +67,11 @@ const Header: React.FC<IHeader> = ({ title, debounceSearch }) => {
     handlePopState();
   }, [location]);
 
+  useEffect(() => {
+    const timer = isFlipping ? setTimeout(() => setIsFlipping(false), 500) : undefined;
+    return () => clearTimeout(timer);
+  }, [isFlipping, setIsFlipping]);
```
Tushar-4781, line 73: @vinaybadgujar102 please correct me if i m wrong
Including setIsFlipping in the dependency array might seem redundant in this context. When setIsFlipping is called, it inherently triggers a change in the isFlipping state right ?

src/store/index.ts
```diff
@@ -95,4 +95,9 @@ export const focusTaskTitle = atom({
   default: "",
 });
 
+export const flipAnimationState = atom({
```
Tushar-4781, line 98: why we are making it a global state ? and not use it as local state?

Tushar-4781 (commented):
Nicely done @vinaybadgujar102 
Can you please resolve my 2 queries before we merge this ?

src/store/index.ts
vinaybadgujar102: if we use local state, the header component re-renders when handlePartner() is called which resets the flipping state to false

src/components/Header/Header.tsx
vinaybadgujar102: agreed, including "setIsFlipping" is redundant

Tushar-4781 (approved):

PR #1682, opened 2023-12-10
fix goal action not opening for sub-goals in partner mode

### LINKED ISSUE #1590: Can't react to goals shared with me
**Describe the bug**
According to https://github.com/tijlleenders/ZinZen/issues/1491 I should be able to make suggestions in my partner's goals.  
I can suggest a subgoal to a partner goal. However - I can only suggest edits to the root goal - not the subGoals.

**To reproduce:**

**Expected behavior**
Able to use the actions menu to suggest changes on subGoals in a shared partner goal.

**Screenshots**

**Additional context**
Probabaly introduced by PR moving edit menu to the goal circle menu.

vinaybadgujar102:
resolves #1590

### DIFF
--- a/src/components/GoalsComponents/MyGoal.tsx
+++ b/src/components/GoalsComponents/MyGoal.tsx
@@ -132,7 +132,7 @@ const MyGoal: React.FC<MyGoalProps> = ({ goal, showActions, setShowActions }) =>
   };
   async function handleDropDown(e: React.MouseEvent<HTMLDivElement, MouseEvent>) {
     e.stopPropagation();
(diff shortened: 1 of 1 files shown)


tijlleenders:
@vinaybadgujar102 
I confirm this is working now - but I do get a weird change every time for the days of the week - even though that wasn't suggested and there doesn't seem to be any difference... Can you check the logic for this?

image

@Tushar-4781  OK to merge?
This PR is about being able to make the suggestion.  
I think I'll file a new issue as the accepted suggestion is not propagated back to the initial suggestor.

Tushar-4781 (approved):

PR #1691, opened 2023-12-15
add hint toggle in config modal

### LINKED ISSUE #1629: Add 'Hints' toggle to Goal and Budget edit screen
It's very nice to get hints.

Add a (default switched off) Hints toggle to the Goal and Budget create/edit menu like so:
image

Integrating the text inside the toggle like so:
image

This does however create a very big area of action color on the screen. It also is less readable/accessible with font size being smaller and lower contrast. 
The alternative with 'Hints' outside of the toggle is better/cleaner/more discreet/more accessible/less noisy.

**Additional context**

Image

Image

The Hints functionality will not be complete - but this toggle is the first step.  
It later has to be connected to querying the pseudoanonymousmessages endpoint to retrieve any hints on a random weekly basis - and create a notification dot if there are new Hints, together with a Hint modal with 'Report'/'Dismiss'/'Next'/'Add'/'Snooze for a week' buttons.  
There will be separate issues for that.

vinaybadgujar102:
resolves #1629

### DIFF
--- a/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.scss
+++ b/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.scss
@@ -129,3 +129,7 @@
   font-size: 17px !important;
 }
 
+.ant-switch {
+  margin-top: 6px;
+}
+
--- a/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.scss
+++ b/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.scss
@@ -140,7 +140,7 @@
   }
 }
 
-.ant-switch {
+.ant-switch, .ant-switch-dark {
   margin-top: 6px;
 }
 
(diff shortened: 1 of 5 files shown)


tijlleenders:
Regression effect:
image

tijlleenders:
Can you align the Hints text left?
image


Also, when not active the button should have a grey color, like so:
image

vinaybadgujar102:
Only hint text should be aligned?

tijlleenders:
I don't understand.  
Align all things to make it look nice.

tijlleenders (changes requested):
Thanks Vinay.

I'm not an expert on UI design - but I know this is not recognizable as a toggle:


We need a better foreground, background and outline color for dark mode.  
You can use the internet as inspiration : 



Also, I think the X is too busy/noisy.  
Could you remove it so the default off setting doesn't "shout"?  
Please leave the checkmark for now.

tijlleenders (approved):
We can use it like this!

@Tushar-4781 OK on code?

Tushar-4781 (approved):

PR #1694, opened 2023-12-16
send list of goals to scehduler instead of map

### LINKED ISSUE #1597: Send goals to scheduler in a list instead of map
It's very annoying when a data structure is more complicated than it needs to be.

Make the input to the scheduler a list, instead of a map. Currently the id's are used as the map key, but each goal also has an id attribute. We can do without the map key, just send a list of Goals.

None.

**Additional context**
Scheduler will break - but that is temporary.

vinaybadgujar102:
resolves #1597


tijlleenders (approved):
LGTM Agree @Tushar-4781 ?

Tushar-4781 (approved):

PR #1717, opened 2023-12-24
created: adr of next js

Tushar-4781:


tijlleenders (approved):
I'm deeply sceptical about how a back-end focused framework will improve an offline-first PWA that has no backend...  
but let's see!

PR #1743, opened 2024-01-06
call hints api when goal item is saved with hints toggle enabled

vinaybadgujar102:
@tijlleenders, implemented the basic required functionality of hitting the API. currently not storing the goal's hint state in db as I need initial review regarding whether the request is sent correctly or not.


vinaybadgujar102:
@Tushar-4781 can we store hints field in goalCollection

vinaybadgujar102:
@tijlleenders, hints field is stored in new "hintsCollection" as suggested by Tushar.

tijlleenders (changes requested):
Yep, this seems to work!

However, when a parent goal has hints disabled - that parent should not get sent in the request - as it should remain private.

Correct that and I'm good to pass on to code review.

vinaybadgujar102:
@tijlleenders made requested changes

tijlleenders (dismissed):
@vinaybadgujar102 This is working correctly now!

@Tushar-4781 OK to merge?
FYI there are two network requests made each time - I don't know if we need the OPTION - but that should be a separate issue if not.

src/models/GoalItem.ts
```diff
@@ -32,4 +32,5 @@ export interface GoalItem {
   } | null;
   typeOfGoal: "myGoal" | "shared";
   newUpdates: boolean;
+  hints: boolean;
```
Tushar-4781, line 35: why do we need hints in goal item ?

src/api/GoalsAPI/index.ts
```diff
@@ -171,6 +174,42 @@ export const shareMyGoalAnonymously = async (goal: GoalItem, parent: string) =>
...
+  let parentGoalHint = false;
+  if (goal.parentGoalId !== "root") {
+    parentGoalTitle = (await getGoal(goal.parentGoalId))?.title || "";
+    parentGoalHint = (await getGoalHintFromDB(goal.parentGoalId)) || false;
+  }
+  const requestBody = {
+    method: "getHints",
+    installId: getInstallId(),
+    parentTitle: parentGoalTitle,
+    goal: {
```
Tushar-4781, line 188: why are we extracting fields here, can't we pass the full goal object. use object deconstruction if you want to discard private fields like participants.

src/api/HintsAPI/index.ts
```diff
@@ -0,0 +1,42 @@
+import { db } from "@src/models";
+
+export const getGoalHintFromDB = async (goalId: string) => {
```
Tushar-4781, line 3: don't use inDb/fromDb the dir is design to contain the methods to perform CRUD ops on DB. so just go with simple naming. :)

Tushar-4781 (changes requested):

src/api/GoalsAPI/index.ts
vinaybadgujar102: @tijlleenders does the request body need the whole goalItem to be passed?

Tushar-4781: not sure

tijlleenders: > @tijlleenders does the request body need the whole goalItem to be passed?

No - just there are only a few goal fields expected in the API:  
- title
- duration (optional)

src/helpers/GoalController.ts
```diff
@@ -32,6 +35,12 @@ export const createGoal = async (
     goalColor,
   });
 
+  if (goalHint === true) {
```
Tushar-4781, line 38: you can directly do if(goalHint)

src/helpers/GoalController.ts
```diff
@@ -62,7 +71,19 @@ export const modifyGoal = async (
   goalTitle: string,
   goalColor: string,
   ancestors: string[],
+  goalHint: boolean,
 ) => {
+  if (goalHint === true) {
+    const res = await getGoalHints({
```
Tushar-4781, line 77: why we are fetching hints here ?

src/helpers/GoalController.ts
```diff
@@ -72,6 +93,7 @@ export const modifyGoal = async (
     goalColor,
   });
   sendUpdatedGoal(goalId, ancestors);
+  await updateHint(goalTags.id, goalHint);
```
Tushar-4781, line 96: can we make it async ?

Tushar-4781 (changes requested):

src/helpers/GoalController.ts
vinaybadgujar102: @Tushar-4781 The function getGoalHints() connects to the API mentioned in the issue when the user turns on the hints toggle and saves a goal. 
There's another function called getGoalHint() that fetches the hint value from the "hintsCollection" for a specific goal. To avoid confusion, we need better names for the getGoalHints() function. Any ideas? @tijlleenders

src/helpers/GoalController.ts
vinaybadgujar102: @Tushar-4781 Which function are you referring to: "sendUpdatedGoal" or "updateHint"?

src/helpers/GoalController.ts
tijlleenders: - getHintsFromAPI()
- getHintsFromCollection()
? The days where we need to use short names to save space are over :)

Tushar-4781: @vinaybadgujar102 I still don't understand properly why we are fetching hints here. I mean you are not using it anywhere so what's the point of fetching if?

src/helpers/GoalController.ts
Tushar-4781: Both

Tushar-4781: @vinaybadgujar102 please make them async

Tushar-4781 (changes requested):

src/helpers/GoalController.ts
Tushar-4781: @vinaybadgujar102 this conversation is not resolved please make them async

vinaybadgujar102: @Tushar-4781 From my understanding, the code is already asynchronous using 'await'. Can you provide more details on the suggested changes?

Tushar-4781: Utilizing await will pause subsequent execution, consequently impeding the progress of goal updates. Employing then and catch for each function allows them to operate asynchronously in the background, as there is no reliance on their outcomes, enabling independent execution in the background

vinaybadgujar102:
```
const sendUpdatedGoalPromise = sendUpdatedGoal(goalId, ancestors);
const updateHintPromise = updateHint(goalTags.id, goalHint);
await Promise.all([sendUpdatedGoalPromise, updateHintPromise]);
```
@Tushar-4781 will this work?

Tushar-4781: @vinaybadgujar102 
1. use allSettled
2. remove the await and use catch for handling the error.

tijlleenders:
@vinaybadgujar102 Please re-request a review from Tushar once done. It's not clear if you're done right now.

src/helpers/GoalController.ts
vinaybadgujar102: @Tushar-4781 I have updated the code, please review

Tushar-4781 (approved):

PR #1758, opened 2024-01-08
scheduler 2024-01-08

tijlleenders:
This scheduler version has a few improvements:  
- shows the first day  
- outputs impossible as a list of ImpossibleTask

Probably some additional commits are required to make frontend be able to handle the new way of displaying impossible Tasks. This is now prohibiting display of the schedule:  
image

@vinaybadgujar102 Can you check this branch out and fix it before we merge?


vinaybadgujar102:
@tijlleenders, fixed the type error

vinaybadgujar102 (approved):

Issue #1796, opened 2024-01-21
Align + to center of action button


tijlleenders:
**Is your feature request related to a problem? Please describe.**
It's very annoying when things are not aligned.

**Describe the solution you'd like**
Align the + to the center of the action button.

**Describe alternatives you've considered**
None.

**Additional context**
![image](https://github.com/tijlleenders/ZinZen/assets/44887755/2008f503-6a11-4bf1-b803-dc40cf7aa9bb)

aabse:
Hi! Can I work on it?

tijlleenders:
Sure thing @aabse !

PR #1798, opened 2024-01-22
Fix: Image aligned in the center inside button

aabse:
This PR fix this issue https://github.com/tijlleenders/ZinZen/issues/1796

image


Tushar-4781 (approved):
Thanks for fixing this @aabse

Issue #1809, opened 2024-01-25
Allow 'enter' on keyboard for saving input on browser edit screen


tijlleenders:
**Is your feature request related to a problem? Please describe.**
It's very annoying when you're not on mobile - and try to save a goal by pressing enter.

**Describe the solution you'd like**
Save the goal when enter is pressed while editing the title.

**Describe alternatives you've considered**
Go to the next input field. That might be better - till you reach the save input and then exit the edit screen. Let's try the direct save first.

**Additional context**
We're currently only developing for mobile - so this is low prio and only useful when testing things quickly not related to the device.

devndevs:
I will be happy to help with this.

devndevs:
@tijlleenders are we good to close this issue out?

tijlleenders:
Hi @devndevs it will be auto-closed when the PR merges - since we mentioned 'resolves #' in the PR description. ;)

tijlleenders:
Oh we didn't - yeah I'll close it.

PR #1810, opened 2024-01-26
[Issue #1809] || Allow 'enter' on keyboard for saving input on browser edit screen

devndevs:
## Description of the Original Issue
As detailed in [Issue #1809](https://github.com/tijlleenders/ZinZen/issues/1809), the problem appears to be related to the budget variation of the goal feature in our application. Specifically, the issue involves the inability to save and close the budget modal when pressing the Enter key, particularly when using a desktop browser.

While the original issue did not explicitly specify that the problem was confined to the budget variation, my investigation led me to this conclusion, and the implemented solution is focused on addressing this aspect.

## Implemented Solution
To resolve this, I wrapped the contents of the modal in a form that allowed me to reduce redundancy by utilizing onSubmit to allow the goal to be saved by using enter.

### DIFF
--- a/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.tsx
+++ b/src/components/GoalsComponents/GoalConfigModal/ConfigGoal.tsx
@@ -301,44 +301,45 @@ const ConfigGoal = ({ goal, action }: { action: "Update" | "Create"; goal: GoalI
                       : `${budgetPerHrSummary} hr / day, ${budgetPerWeekSummary} hrs / week`,
                     body: (
                       <div>
-                        <span>{budgetPerHrSummary} hrs / day</span>
-                        <Slider
-                          tooltip={{ prefixCls: "per-day-tooltip" }}
-                          min={0}
-                          max={beforeTime - afterTime}
-                          marks={{
-                            0: "0",
-                            [perDayHrs[0]]: `${perDayHrs[0]}`,
-                            [perDayHrs[1]]: `${perDayHrs[1]}`,
-                            [beforeTime - afterTime]: `${beforeTime - afterTime}`,
-                          }}
-                          range
-                          value={perDayHrs}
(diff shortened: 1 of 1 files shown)


tijlleenders (changes requested):
Wow that's an excellent PR description @devndevs !

About the code:  
Can't we just remove the 'if Goal/Budget and go straight to `await handleSave()` on Enter?

Tushar-4781 (changes requested):
hi @devndevs we are planning to implement form so that it auto captures the submission. via on submit. do you think you can wrap this component in form. it will save repeating LOC for us. for example "e.preventDefault(); await handleSave()"

devndevs:
> hi @devndevs we are planning to implement form so that it auto captures the submission. via on submit. do you think you can wrap this component in form. it will save repeating LOC for us. for example "e.preventDefault(); await handleSave()"

I have done what you have requested. Wrapped the contents of the modal in a form which allowed me to reduce redundancy by utilizing on submit to allow the goal to be saved on using enter.

tijlleenders (changes requested):
@devndevs Thanks for all this effort!

I can confirm it's working now on Firefox and Chrome...
... except when adding a new Goal.  

When editing an existing Goal it works fine.  
Also when adding as well as editing a Budget.  

Could you maybe investigate?

devndevs:
> @devndevs Thanks for all this effort!
> 
> I can confirm it's working now on Firefox and Chrome... ... except when adding a new Goal.
> 
> When editing an existing Goal it works fine. Also when adding as well as editing a Budget.
> 
> Could you maybe investigate?

Hi there @tijlleenders!

May I have a little more context on the issue you are seeing? I am testing it and it is working on my end.

tijlleenders (approved):
Sorry @devndevs it was a caching issue...

All good here, thanks for the great work!

Tushar-4781 (approved):
Thanks 👍

Issue #1531, opened 2023-10-03
Quick theme toggle button disappears from title bar 


Shubham-Praj:
**WON'T FIX - Not a bug in my view.**
If you believe this is a bug - please upvote this issue.

**Describe the bug**
When the user switches from light theme to dark theme using the theme toggle button in title bar, the toggle button disappears in dark mode

**To Reproduce**
When present on light mode, toggle the theme to dark mode

**Expected behavior**
The quick theme toggle button should be present on dark mode as well.
When present in dark mode the quick theme toggle button should show a bright bulb icon (which on click toggles to light mode)

**Screenshots**
Light theme 
![image](https://github.com/tijlleenders/ZinZen/assets/62180559/ddbb271d-6156-48ee-831f-5758c74ffee8)

Dark theme
![image](https://github.com/tijlleenders/ZinZen/assets/62180559/8dbf6e81-6bf8-4510-a528-a113336eb111)

tijlleenders:
@Shubham-Praj Thanks for reporting this.

I understand how this can be confusing. The quick switch is time dependant. It obly appears for dark when dark mode is probably preferable and vice-versa - see #1380.

So not a bug - it's a feature...
Leaving the quick-switch icon permanently wpuld crowd the header.

... but I'll leave this one open so people who find this confusing as well can vote to change this behavior - or propose a better solution.

devndevs:
I appreciate the intention behind the feature designed to declutter the header, though it might be a bit perplexing for users not yet acquainted with this functionality or its rationale.

For instance, when I selected the "Light Mode" icon this morning, I was initially taken aback by its disappearance. However, I managed to locate the toggle in the menu. This experience, albeit minor, highlighted an opportunity to enhance the UI's intuitiveness.

It's also worth noting that a significant number of users—over 80% according to several surveys—express a preference for Dark Mode, with some findings even suggesting a preference as high as 94%. This is an important consideration, especially for those moments when users, like myself, are navigating to adjust their settings in the dropdown menu. Ensuring the toggle's constant accessibility could greatly improve user satisfaction.

In light of this, I'd like to propose a couple of constructive suggestions:

1. If the objective is to minimize header clutter, perhaps re-evaluating the necessity of the quick toggle could be beneficial. It seems that in many contemporary applications, the Dark Mode setting is tucked away within the settings menu, reserved for one-time adjustments rather than frequent toggling. This approach could simplify user interactions and streamline the overall experience.
[...]

PR #1819, opened 2024-01-31
[Issue #1531] || fix darkmode toggle bug

devndevs:
Description of the Original Issue
As detailed in https://github.com/tijlleenders/ZinZen/issues/1531, the problem appears to be related to the quick toggle feature to switch between dark and light modes. The Icon would disappear depending on what time it was, in the case of the example shown it was when the nighttime logic was active, as I tested it the nighttime logic was not active so the opposite was happening to me  (absence of dark mode icon)

Implemented Solution
I changed the logic to show the opposite icon regardless of what time it was on the user's computer.

### DIFF
--- a/src/components/Header/Header.tsx
+++ b/src/components/Header/Header.tsx
@@ -113,7 +113,7 @@ const Header: React.FC<IHeader> = ({ title, debounceSearch }) => {
             {isNighttime || darkModeStatus ? (
               <HeaderBtn
                 path={darkModeStatus ? lightModeIcon : darkModeIcon}
-                alt={darkModeStatus ? "light mode" : "dark mode"}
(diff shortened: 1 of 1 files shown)


tijlleenders:
So sorry @devndevs I should have marked the issue as "won't fix".  
The comments say it's not a bug. If you believe it is - please vote on the issue.

tijlleenders (approved):
As @devndevs suggested I'll make an issue to remove the quick toggle so we can declutter the header - and make dark mode the default.

tijlleenders:
For now we can use this. I'll pay attention to my own needs - I usually only use dark mode - so based on n=1 @devndevs right :)

Tushar-4781 (approved):
Thanks for fixing this @devndevs

PR #1821, opened 2024-02-02
display a visual indicator "!" (exclaimation mark) for goals that the scheduler has marked as impossible

### LINKED ISSUE #1377: Show impossible on Goal and Goal parents
It's very annoying when something is impossible and you only find out when it's too late.

Whenever there is a Goal that:  
- has impossible Tasks from scheduler output
show a ! inside the circle of the corresponding Goal, and any parents.  
When navigating 'down' the Goal hierarchy, always order Goals with a ! on top - so you can easily find the Goal that has an issue quickly.

vinaybadgujar102:
resolves #1377


tijlleenders:
Thanks @vinaybadgujar102 

Good you check in before going too far.


The MyTime should not show anything ab9ut impossible.

On the MyGoals - please remove the tiny circle with exclamation - and just prepend the title with a "! " - we'll figure out an appropriate design at a later stage.

tijlleenders:
@vinaybadgujar102 Sorry I missed this one earlier this week.  

This is a great first step. Of course - it's not super useful to just have an exclamation mark - withouth having a way to know more about what the problem is exactly, and how to fix it...  

That is OK for now.  
This is already a step up from not knowing there is a problem at all. The idea is that we can at some point prevent most problems by detecting them before you can press save in the edit menu.  


@Tushar-4781 OK on code?

tijlleenders (approved):

src/hooks/useIsGoalImpossible.tsx
```diff
@@ -0,0 +1,34 @@
+import { getChildrenGoals } from "@src/api/GoalsAPI";
+import { getImpossibleGoalById } from "@src/api/ImpossibleGoalsApi";
+import { useEffect, useState } from "react";
+
+const useIsGoalImpossible = ({ id }: { id: string }) => {
```
Tushar-4781, line 5: this is a little too costly. we can't create a hook that runs for ever goal every time a single change in DOM will trigger this. can we think of some better approach ?

src/components/MyTimeComponents/MyTimeline/updateImpossibleGoals.ts
```diff
@@ -0,0 +1,36 @@
+import { getGoal } from "@src/api/GoalsAPI";
+import { deleteAllImpossibleGoals, addImpossibleGoal } from "@src/api/ImpossibleGoalsApi";
+
+interface FormattedGoal {
+  goalId: string;
+  goalTitle: string;
+}
+
+const fetchAndFormatGoal = async (taskId: string): Promise<FormattedGoal | null> => {
+  try {
```
Tushar-4781, line 10: try catch is of no use here. instead you can place it in getGoal api

Tushar-4781 (changes requested):

src/hooks/useIsGoalImpossible.tsx
vinaybadgujar102: Hi @Tushar-4781, I am unable to find an alternative solution. Could you share any ideas or suggestions?

Tushar-4781: i think this involves some refactoring in components. can you resolve failing tests of #1837 cuz i did some refactoring already. post that we can think of a better structure cuz it little difficult for me to sync in.

tijlleenders:
@vinaybadgujar102 What is needed to get this merged?  
The PR has been hanging around for too long.

src/hooks/useIsGoalImpossible.tsx
vinaybadgujar102: @Tushar-4781, currently we are passing goals as props through multiple components. As a first step toward refactoring can we store the fetched goals in the recoil state and then directly use that state in the required component?

vinaybadgujar102:
@tijlleenders, as per the review by Tushar, there is a need for some refactoring in the code. Can we discuss the refactoring points in a new PR?

tijlleenders:
> @tijlleenders, as per the review by Tushar, there is a need for some refactoring in the code. Can we discuss the refactoring points in a new issue?

That's a question for @Tushar-4781 .

Anyway, there are merge conflicts that need to be resolved too.

tijlleenders:
> > @tijlleenders, as per the review by Tushar, there is a need for some refactoring in the code. Can we discuss the refactoring points in a new issue?
> 
> That's a question for @Tushar-4781 .
> 
> Anyway, there are merge conflicts that need to be resolved too.

@Tushar-4781 *Bump*

Tushar-4781:
sorry for the late response, couldn't find time for this one. quick solution would be to fetch the impossible goal(s) and set it in a global state and then we can use it in all components.

vinaybadgujar102:
@Tushar-4781 
- Created a state for impossible goals
- Fetch and store them when "My time" is visited
- In the GoalList component, updated the goals to include the impossible field.
- passed those updated goals to GoalList

tijlleenders:
Looks good.
Still need to go to Mytime to trigger a new schedule ... but real-time feedback is a separate feature to think about.

tijlleenders:
@Tushar-4781 Code OK?

src/api/ImpossibleGoalsApi/index.ts
```diff
@@ -0,0 +1,54 @@
+import { db } from "@src/models";
+import { getGoal } from "../GoalsAPI";
+
+export const addImpossibleGoal = async (impossibleGoalId: string) => {
+  try {
+    const goal = await getGoal(impossibleGoalId);
+    if (!goal) throw new Error("Goal not found.");
```
Tushar-4781, line 7: we can't throw error directly like this as it will break the code instead a custom hook can be created that catches error. can you create a custom hook that can be used as a wrapper for all api(s)?

src/Interfaces/index.ts
```diff
@@ -31,3 +31,7 @@ export interface ILocationState {
   displayFocus: boolean; // whether to show the focus component
   allowAddingBudgetGoal?: boolean; // whether to allow adding a budget goal
 }
+
+export interface GoalWithImpossible extends GoalItem {
```
Tushar-4781, line 35: can we name it ImpossibleGoal instead of GoalWithImpossible

Tushar-4781 (changes requested):

src/api/ImpossibleGoalsApi/index.ts
vinaybadgujar102: Could you please provide some suggestions on how we can create the custom hook you mentioned? I have experience with creating wrapper functions for HTTP API calls but can't come up with logic for doing the same for our API, which directly communicates with the database.

Tushar-4781: the hook will just be a functional component that takes the function in argument and will call it and if that fn throws an error it will catch it & will show the same in toast.

vinaybadgujar102: @Tushar-4781, I tried but couldn't develop the hook that can work for all our current APIs. For now, we can console log this errors instead of throwing it. We can take care of your concern in the new PR.

tijlleenders (approved):

tijlleenders:
@vinaybadgujar102 Can you make an issue for the error hook? Then we can merge this.

PR #1826, opened 2024-02-05
edit: set dark theme as default for new users

### LINKED ISSUE #1820: Set dark theme as default for new users
It's very annoying when most users prefer a dark theme and the default is light.

Set dark theme as default for new users - from language choice onward.

None. We have to choose a default. Dark seems to be more in demand. You can always change it via settings.

**Additional context**
Thanks @devndevs for the suggestion.

vinaybadgujar102:
resolves #1820 

- modified getDarkModeValue() to return "on" by default
- modified useApp() hook to set "darkMode" key in localStorage to "on"

### DIFF
--- a/cypress/integration/bottom-nav.spec.ts
+++ b/cypress/integration/bottom-nav.spec.ts
@@ -30,14 +30,14 @@ describe("BottomNavbar", () => {
 
   it("should change to next theme when Next button is clicked", () => {
     cy.get(".bottom-nav-item").contains("Next").click();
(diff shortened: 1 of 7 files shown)


tijlleenders (changes requested):
Can't read the question on the language choice page.

tijlleenders:
This is not dark mode compatible: 
image

tijlleenders (approved):
Yeah, this works.

@Tushar-4781 Code OK?

Tushar-4781 (approved):

PR #1837, opened 2024-02-17
New Feature: Soft Delete Goal

### LINKED ISSUE #1625: Soft delete
It's very annoying when you accidentally press Delete on a Goal, especially when it contains a whole tree of things. Right now this is pretty easy to do - because the edit menu is 25% reserved for the Delete button.

Make the Delete a soft delete. Whenever you delete something, it gets placed in the Archive for 7 days before it is actually deleted. 

Make the delete button smaller. This would reduce the risk, but it can still happen.

**Additional context**
No need to differentiate between the real archived goals and the soft-deleted goals. If needed we can make a separate issue for that.

Tushar-4781:
This PR introduce soft delete functionality which works at all levels and in sharing.
Edge Case Left - If a top (root) level goal is deleted which is colab goal then the current implementation will not give option to restore it.

Resolves #1625


tijlleenders:
@vinaybadgujar102 Please investigate the  failing tests so we can get this merged.

tijlleenders (approved):
Awesome!

Tushar-4781:
@tijlleenders @vinaybadgujar102 i have zero clue how its passing the test in the PR. i ran locally still tests are failing i.e. 1 failed 4 skipped
image

tijlleenders (changes requested):
Investigate tests failing on merge.

PR #1842, opened 2024-03-01
solved can't create a goal inside a budget 

### LINKED ISSUE #1839: Can't create a Goal inside a Budget
**Describe the bug**
When I try to add a Goal to a Budget - it shows as a Budget.

**To Reproduce**
2024-02_cant_create_goal_within_a_budget

**Expected behavior**
A Goal of 1h should be shown inside 'Work' Budget.

**Additional context**
None.

vinaybadgujar102:
resolves #1839
### Problem 
- When users make a new goal or budget inside another goal or budget without filling in certain details, the new goal or budget would just copy those details from the bigger goal or budget inside.
- At first, it was okay for a new goal or budget to copy details like deadlines and time budgets from the one above it.
- Later, we made two separate things: goals and budgets. Goals needed certain details like the due date and duration but didn't care about other details like time budgets, after time, before time.
- But, when you made a goal inside a budget, it would still copy all the details from the budget, even the ones that didn't make sense for a goal. This made the goal seem like a budget, which we didn't want.

### Solution
- Don't inherit budget-related properties as they are always automatically added during budget creation.

### DIFF
--- a/src/utils/index.ts
+++ b/src/utils/index.ts
@@ -125,9 +125,6 @@ export function inheritParentProps(newGoal: GoalItem, parentGoal: GoalItem) {
   if (!goal.habit) {
     goal.habit = parentGoal.habit;
   }
-  if (!goal.on) {
-    goal.on = parentGoal.on;
-  }
 
   goal.rootGoalId = parentGoal.rootGoalId;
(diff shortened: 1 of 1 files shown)


tijlleenders (approved):
Awesome!

@Tushar-4781 Agree on code?

PR #1846, opened 2024-03-16
restrict property inheritance by subgoals

### LINKED ISSUE #1815: Restrict Goal property inheritance by subGoals to color and language
It's very annoying when SimpleGoals and BudgetGoals get their properties mixed. Currently this is happening where BudgetGoal-specific fields like `on`, `filters` and `timeBudget` get inherited by SimpleGoals that are subGoals of that BudgetGoal.

Only let subGoals inherit the color of the parent goal. The type (Simple/Goal) of parent or child don't matter.

The language of the created Goal is the currently selected language of the interface.

Make two different types. That would make UI harder as for navigation/etc it has benefits to treat them as the same type. Maybe that can be solved by having two types with a same 'hierarchy' interface - but that's beyond this issue. This is a quick fix.

**Additional context**
None.

vinaybadgujar102:
resolves #1815


tijlleenders (approved):
LGTM!

@Tushar-4781  Agree on code?

tijlleenders:
@vinaybadgujar102 Can you solve the merge conflicts?

tijlleenders (approved):

PR #1847, opened 2024-03-17
restrict min-max weekly hours based on daily budget hours

### LINKED ISSUE #1783: Enforce consistency in budget
It's very annoying when you can define an unrealistic budget.

Avoid that you can make a budget in which:  
- the sum of all max/day is not lower than the min per week
  For example, if I have a goal of exercising 1-2 hours per day...
  ... I should not be able to set a minimum per week of more than 14h.
- sum of all min/day is not lower than the min per week - or the max per week
  For example, if I have a goal of exercising 1-2 hours per day...
  ... I should not be able to set a min per week or a max per week lower than 7  

None.

**Additional context**
Scheduler will panic if this happens - so best to avoid it on the input side (edit budget screen).  
This could require adapting the budgets of the default goals so that some of them have 0 as the min per day.

vinaybadgujar102:
resolves #1783


tijlleenders (changes requested):
Pretty awesome @vinaybadgujar102 

I did however (not clear how to reproduce) get into an invalid state:  


Another thing is that the number of days is always assumed to be seven. Not sure if that can be taken into account in this issue or if we shoukd make a separate issue for that.

vinaybadgujar102:
@tijlleenders, please check now to see if the mentioned concern has been resolved

vinaybadgujar102:
@tijlleenders made the requested changes, please review

tijlleenders (changes requested):
The budget is still unresponsive to changes in days of the week:
2024-03-29 budget unresponsive to days of the week

tijlleenders (changes requested):
Perfect!

Only... after saving / close / open thr preview is not visible anymore:

vinaybadgujar102:
@tijlleenders, fixed

tijlleenders (dismissed):
Super!

@Tushar-4781 Code OK?

src/components/GoalsComponents/GoalConfigModal/ConfigGoal/BudgetAccordian/BudgetDaySlider.tsx
Tushar-4781, line 1: i don't think its necessary to create every single element as seperate component unless it is used by multiple components.  not a good move for this 16 LOC. plz merge these components of config goal

Tushar-4781 (changes requested):

src/components/GoalsComponents/GoalConfigModal/ConfigGoal/BudgetAccordian/BudgetDaySlider.tsx
vinaybadgujar102: So should I merge all the separated components or only the mentioned one "DailyBudgetSlider"?

Tushar-4781: merge all of them

tijlleenders (approved):

src/components/GoalsComponents/GoalConfigModal/ConfigGoal/BudgetAccordian/BudgetDaySlider.tsx
vinaybadgujar102: Done

tijlleenders (changes requested):
Found a last minute glitch:

tijlleenders:
Seems this only happens when you don't expand the budget menu - and then try to add / remove days. Adding also doesn't go above the initial amount, even though all days may be selected.

vinaybadgujar102:
The week budget hour value is correctly applied in mentioned screenshot. It is between the constraints as per the condition:
Min: 24 Max: 40. 

If the applied weekly budget values fails to satisfy both min-max constraints then it resolves to single value (x hrs) accordingly, thats why no range (x hrs - x hrs) is displayed in summary.

Same happens with daily budget hours when between slider is changed.

tijlleenders (approved):
@vinaybadgujar102 You're right - somehow as a user I expect things to change when I remove a whole day... but let's try this out for a whioe and see how this goes in real life.

PR #1856, opened 2024-03-22
implement add-filler and fix overflow bug

tijlleenders:
Implemented simple filler goals/activities and fixed an overflow issue.


tijlleenders:
@vinaybadgujar102 Please investigate why ESLint is failing:  
image

Why is eslint not found - when this worked fine before?

vinaybadgujar102 (approved):

PR #1861, opened 2024-03-24
update scheduler

tijlleenders:
When Goals have no deadline, they are no longer considered impossible if the associated Activities can't be scheduled in the requested calendar window.  

Instead, they will be silently ignored - assuming they will be possible after the calendar end.  

This is of course not a perfect assumption - but good enough. The solution to this uncertainty is to extend the calendar window far enough into the future so that any impossible deadline will be noticed in time to adjust expectations/priorities.


tijlleenders:
@vinaybadgujar102 Do you know why ESLint is failing?  
It can't be me, right? All I did was update the scheduler.

tijlleenders:
@vinaybadgujar102 

Thanks. I need a little more context though: Why was this added and why is it OK to remove it now?

vinaybadgujar102:
@tijlleenders eslint removed this configuration option.

https://stackoverflow.com/questions/65675771/eslint-couldnt-find-the-config-prettier-typescript-eslint-after-relocating

tijlleenders:
@vinaybadgujar102 The link you provided seems to only mention `prettier`, not `parser`.  

More importantly:  
What did parser do in terms of checking? (The npm package info)[https://www.npmjs.com/package/@typescript-eslint/parser] indicates this allows ESLint to check TypeScript. Is that all it's for?  

Is that functionality still being provide in some other way - or do we need to replace it instead of simply removing it?

tijlleenders:
Also, just because the check is green doesn't mean there is no problem:  
image

vinaybadgujar102:
> What did parser do in terms of checking? [The npm package info](https://www.npmjs.com/package/@typescript-eslint/parser) indicates this allows ESLint to check TypeScript. Is that all it's for?

Yes, it allows for typescript checking. I referred the docs and also checked for the same error in eslint org's discord. It mentioned that configuration for "typescript-eslint/parser" package should be as the [9c2f32a](https://github.com/tijlleenders/ZinZen/pull/1861/commits/9c2f32ac0fd433cb99317fae5aa3c8a22fdbac84). 

The same functionality of ts checking is provided by "plugin:@typescript-eslint/recommended"

vinaybadgujar102:
I think the eslint workflow is running correctly now. @tijlleenders please verify

tijlleenders:
@vinaybadgujar102 Where is it running correctly? Not in CI:  



A part from that - downgrading from v14 to v1 doesn't seem like a future-proof fix...

vinaybadgujar102:
i mean that workflow is correctly showing lint errors now. the eslint workflow was not correctly running from quite a time I think. You can refer to previous PRs from 1-2 months also. The lint errors can be solved in new PR according to me.

> A part from that - downgrading from v14 to v1 doesn't seem like a future-proof fix...

the version was 1 only, the 4  in 14 accidentally got added in this commit [537a7e2](https://github.com/tijlleenders/ZinZen/pull/1861/commits/537a7e210bd1e1c17e3677fd5f224c73ec23a3e8)

tijlleenders:
Allright, well good to hear you got it sorted @vinaybadgujar102 

The auto check in PR files changed overview is also nice :)  
image

vinaybadgujar102 (approved):

PR #1873, opened 2024-03-31
Update:  Category Attribute for Goals

### LINKED ISSUE #1857: Enforce invariant 'No Budget possible if any ancestor is a simple Goal with duration'
For scheduling, it's very annoying when there is too much complexity. Making a budget inside a simple goal with a fixed duration is very confusing: 
- If the simple goal also has a repeat, it's basically a budget - and should be treated as such even though on the surface it looks like a simple goal. For example the child budget should then minimally conform to the 'simple goal' budget. What??
- If the simple goal doesn't have a repeat, then what does the child budget do? Does it automatically stop when the hours are exhausted? And what if there are sub-budget? Then the simple Goal has repercussions multiple levels down on the behavior of budgets... And those could then contain simple goals too - which can contain (sub)budget goals...etc...

So, the conclusion is 'Disallow Budgets if any ancestor is a simple Goal with duration'.  
This means:  
- don't show the Budget creation option after tapping + if any parent/ancestor is a simple goal WITH duration - just jump straight to the create Goal screen.
- show the Budget creation option after tapping + if there are parents/ancestors - but none of them have a duration
- disallow adding a duration to a simple Goal if any of the children/descendants are a Budget

See problem statement... The alternative is very complex.

**Additional context**
None.

### LINKED ISSUE #1858: Enforce invariant 'Start-Deadline always within parent/ancestor Start-Deadline'
It's very annoying when an invalid scheduling tree is constructed. This messes up the Calendar and requires the input Goals to be mutable - or additional logic to be added in the scheduler to 'correct' if correction is even possible. There is no good answer to an invalid tree. Fixing the error hides the issue from me - and not fixing the issue gives me an error / no Calendar.  

The only right solution is to prevent an invalid scheduling tree in the first place through UI create/update checks:  
Enforce that no simple Goal can have:  
- a Start datetime that is before any of the parent/ancestors
- a Deadline datetime that is after any of the parent/ancestors  

This should already be the case since Goals are not (yet) allowed to be moved around, and Start can't be explicitly set (deduced: start of day of the create Date).  

For deadline anything after the nearest deadline of any Ancestors should be greyed out in the Deadline selection calendar widget.

See problem statement.

Tushar-4781:
Resolves #1857


tijlleenders:
@Tushar-4781 This works as expected!

The only thing is that when trying to create a goal inside a goal with duration - but then you click outside the modal, the navigation goes back up again one level:
2024-03-31 Navigation goes up on cancel new goal creation

tijlleenders:
@Tushar-4781 
Also, it seems like the ancestor check only goes up one level.
If I place an 'placeholder' goal without duration below a goal with duration - I am then allowed to make a budget inside of that - which shouldn't be allowed:


https://github.com/tijlleenders/ZinZen/assets/44887755/b5bda16c-530c-4a12-8ec9-fca5724903c6

tijlleenders (changes requested):
Is #1858 also fixed?

I'm able to set a deadline of a child past the deadline of an ancestor.

tijlleenders (dismissed):

tijlleenders:
Sure @Tushar-4781 !

Can we fix the ESLint though?
I'd like to keep the main branch green from now on, preferably without hardcoding ESLint ignores :).

Tushar-4781:
> Sure @Tushar-4781 !
> 
> Can we fix the ESLint though? I'd like to keep the main branch green from now on, preferably without hardcoding ESLint ignores :).

i don't see these warning locally. will try to fix it some other way

tijlleenders (approved):

PR #1890, opened 2024-04-18
Enhance Hint Management to Exclude Deleted Hints

### LINKED ISSUE #1885: Forget a hint
It's very annoying when you never want to see a certain hint again and it keeps coming back.

Provide a 'Forget hint' option. This is different from 'Report hint' as it validates the hint as being OK / not-spam - it's just not for you.

None.

vinaybadgujar102:
resolves #1885  

This Pull Request introduces crucial updates to our hint management system, ensuring that once a hint is deleted by the user, it does not reappear in subsequent sessions or updates. The changes include:

- **Modification of deleteGoalHint function:** Now adds deleted hints to a dedicated deletedGoalHints array, preventing them from being mistakenly re-added.

- **Modification of updateHintItem function:** Implements checks to filter out any hints already marked as deleted based on their titles before processing updates.


tijlleenders (approved):

tijlleenders:
@Tushar-4781 Code OK?

src/api/HintsAPI/index.ts
```diff
@@ -44,22 +44,37 @@ export const addHintItem = async (goalId: string, hint: boolean, goalHints: IGoa
  * @param {string} goalId - The ID of the goal to update the hint for.
  * @param {boolean} hint - The new hint value to update.
```
Tushar-4781, line 45: why are we adding these descriptions and comments here. we are using Typescript not pure JS. interface and types can be created you can add comments there **if reqd**. otherwise type, name are already defined which is understood. we are adding unnecessary code which is not required in these scenarios.

src/models/HintItem.ts
```diff
@@ -9,6 +9,7 @@ export interface HintItem {
   id: string;
   hint: boolean;
   goalHints: IGoalHint[];
+  deletedGoalHints?: Omit<IGoalHint, "id">[];
```
Tushar-4781, line 12: shouldn't we just store id here ?

vinaybadgujar102: the response we get from the api is an hint goal object with title, parentTitle and duration? fields. We add id to it when we store the hint goal in collection. Thats why used title identify the hint goal

Tushar-4781: @vinaybadgujar102 i was referring to deletedGoalHints field.

vinaybadgujar102: i was referring to same line only. The backend doesn't respond with id for hintGoal that's why we cant uniquely identify it. Thats why we are comparing with title. The id gets added in frontend only.

@tijlleenders can we change the hints API route to store id of each hint

tijlleenders: No the API doesn't store hint IDs as they are anonymously shared amongst all users.

tijlleenders: Once you add a Hint it should get an ID (generated) though - and you can use that normally in the deleted collection.

vinaybadgujar102: > Once you add a Hint it should get an ID (generated) though - and you can use that normally in the deleted collection.

Add hint where? locally or server?

tijlleenders: When you add the suggested hint (without uuid) to your local MyGoals collection - it will get an autogenerated uuid, right?

vinaybadgujar102: yes, id will get added locally. But if we toggle the hints option to off, then we delete the hints array stored locally for that particular goal. When we turn hints back on again, we again add new ids and then store that hints array. So the id gets changed and thus we cant track which hint was deleted if we compare with ids.

tijlleenders:
@vinaybadgujar102 What is the next step to get this PR along?

vinaybadgujar102:
> @vinaybadgujar102 What is the next step to get this PR along?

The feature is fully implemented. I resolved the merge conflicts. Just confirmation from Tushar is pending.

tijlleenders:
@Tushar-4781 Code OK?

tijlleenders:
@Tushar-4781 *bump*

Tushar-4781 (approved):
not 100% aware of the hints flow so far it looks okay.

PR #1907, opened 2024-05-02
place edit modal above keyboard in mobile devices

### LINKED ISSUE #1882: Goal edit modal above keyboard
It's very annoying when a new user tries the Goal create/edit modal but doesn't see the duration and due date fields.

Position the bottom of the Goal create/edit modal relative to the top of the keyboard.

None.

**Additional context**
On older tiny screens or screens with large font size this might hide the title. Those are edge cases and I have no good solution for phones set that way...

On Vanadium there is so much space between the goal edit modal and the keyboard I can't see the title. Best is to minimize the open space between bottom of the Goal modal and the keyboard.

vinaybadgujar102:
resolves #1882

- Built custom hook to detect whether virtual keyboard is active or not. It compares the initial window viewport height with changed one and returns boolean.
- Then applied styles on goal edit modal based on above boolean value.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
         "recoil": "^0.7.1",
         "regenerator-runtime": "^0.13.9",
         "sass": "^1.50.0",
+        "use-detect-keyboard-open": "^0.4.0",
         "uuidv4": "^6.2.13",
         "vite": "^2.9.18",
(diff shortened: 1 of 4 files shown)


tijlleenders (approved):

tijlleenders:
@Tushar-4781 Code OK?

PR #1909, opened 2024-05-03
fix: lint script not working locally

vinaybadgujar102:
The earlier `npm run lint` script was not working properly. Fixed it by updating the command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "preview": "vite preview",
     "test": "jest --verbose ./src/__tests__",
     "cypresstest": "cypress run",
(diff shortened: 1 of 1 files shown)


tijlleenders (approved):

Tushar-4781:
We need to take of warnings as well. like for example unused variables or props etc, issues like this comes under warnings cuz they are not stopping the code from running.  let's remove that flag

vinaybadgujar102:
@Tushar-4781 removed the "quiet" flag as mentioned

Tushar-4781 (approved):

PR #1910, opened 2024-05-06
reschedule feature

### LINKED ISSUE #1400: Develop the reschedule feature for timeline
Reschedule implementation could be as simple as providing a `notOn` for the required 'push' period.  
So reschedule could show you an list of reschedule/postpone options:  
- 1 hour
- 3 hours
- 1 day
- 3 days
- 7 days
- 30 days

If the Goal has a deadline, only show the postpone options that leave at least the duration of the goal between the new start and the deadline.

All the scheduler needs is an extra `notOn` field (separate from filters/timeBudget) for the start of the calendar till the current scheduled start time + postpone time. This field contains **an array** of start/end tuples that represent hours it should not be scheduled on:

```
  {
    "startDate": "2022-01-01T00:00:00",
    "endDate": "2022-01-02T00:00:00",
    "goals": [
      {
        "id": "1",
        "title": "shopping",
        "minDuration": 1,
        "start": "2022-01-01T10:00:00",
        "deadline": "2022-01-01T13:00:00",
        "notOn": [
          {
            "start": "2022-01-01T00:00:00",
            "end": "2022-01-01T11:00:00"
          }
        ]
      }
    ]
  }
```

vinaybadgujar102:
resolves #1400


vinaybadgujar102:
@tijlleenders, updated the notOn's start to be task's start dateTime

tijlleenders:
@vinaybadgujar102 It works out now :)

Can you add a test for this as part of the PR?

Also, this wasn't in the issue description - but added after the first experience: please
- adjust font size of goal title like on MyGoals
- add "Postpone: " to the start of the title
- make the menu selection similar to edit Goal - choose icons as you like but also have the subtext
- no need for a reschedule button - effectuate the postpone as soon as you press one of the options

@Tushar-4781 Code OK?

tijlleenders (changes requested):
Nice!
See comments for a small touchup.

src/components/MyTimeComponents/Reschedule/Reschedule.scss
```diff
@@ -1,108 +1,60 @@
...
-
-  .slot-batch {
-    display: flex;
-    flex-direction: column;
-    gap: 10px;
-
-    div {
-      display: flex;
-      gap: 10px;
+    .ant-radio-group-solid .ant-radio-button-wrapper-checked:not(.ant-radio-button-wrapper-disabled) {
```
Tushar-4781, line 27: instead of targeting this, go for this ".ant-radio-button-wrapper-checked""

src/components/MyTimeComponents/Reschedule/Reschedule.scss
```diff
@@ -1,108 +1,60 @@
...
-      font-size: 0.5em;
-    }
+  .action-btn {
+    color: white;
   }
 
-  .slot-card.selected {
-    font-weight: 600;
+  .radio-button {
+    background-color: rgba(255, 255, 255, 0.08);
```
Tushar-4781, line 41: use a color from the pallet

src/components/MyTimeComponents/Reschedule/Reschedule.scss
```diff
@@ -1,108 +1,60 @@
...
   .action-btn {
-    margin-top: 16px;
-    background: var(--primary-background);
+    color: black;
   }
 
-  div:has(.action-btn) {
-    text-align: center;
+  .ant-radio-group-solid
+    :where(.css-dev-only-do-not-override-w8mnev).ant-radio-button-wrapper-checked:not(
```
Tushar-4781, line 51: cut it short

Tushar-4781 (changes requested):
Can u fix the reschedule popup in light mode.

vinaybadgujar102:
> make the menu selection similar to edit Goal - choose icons as you like but also have the subtext

@tijlleenders please share ss of the edit menu mentioned above

src/components/MyTimeComponents/Reschedule/Reschedule.scss
vinaybadgujar102: > no need for a reschedule button - effectuate the postpone as soon as you press one of the options

as @tijlleenders mentioned to reschedule as soon as user presses the option, we don't need checked based styles anymore

tijlleenders:
@vinaybadgujar102 What I mean is try re-using the already established pattern for a popup with choices - so the user interface is consistent and thus easier to use:
image

tijlleenders (changes requested):
@vinaybadgujar102 Yes, this is awesome!

I see the postpone is only storing one 'notOn' slot?

When I postpone 'Me time' for one hour it moves from 5 to 7 'o clock. When I postpone it again for 1 hour it moves back to 5.

Probably there is only one value being stored in 'notOn' when it's supposed to be an array?

vinaybadgujar102:
> I see the postpone is only storing one 'notOn' slot?
> When I postpone 'Me time' for one hour it moves from 5 to 7 'o clock. When I postpone it again for 1 hour it moves back to 5.
> Probably there is only one value being stored in 'notOn' when it's supposed to be an array?

@tijlleenders fixed

tijlleenders (dismissed):
We're getting there!

vinaybadgujar102:
@Tushar-4781 bump

Tushar-4781 (changes requested):

src/store/TaskState.ts
```diff
@@ -3,5 +3,5 @@ import { atom } from "recoil";
 
 export const displayReschedule = atom({
   key: "displayReschedule",
-  default: null as ITask | null,
+  default: {} as ITask | null,
```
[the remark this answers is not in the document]
vinaybadgujar102: @Tushar-4781 fixed

tijlleenders (approved):
Awesome!

Tushar-4781 (approved):

PR #1915, opened 2024-05-16
fix "completedTodayTimings" is undefined

### LINKED ISSUE #1913: Undefined var
**Describe the bug**
When running dev branch at https://github.com/tijlleenders/ZinZen/commit/b68c060d40e5608d03601dea4f0ad73c94c3622a I get the following error with my own data: 
image

It works fine on production.

**To Reproduce**
Not easy to reproduce without sharing all of my data. 
@vinaybadgujar102 Any idea where this is going wrong?  

The problem is definitely on the UI side, I've run the scheduler with my data - and it produces correct output.

vinaybadgujar102:
resolves #1913 
@tijlleenders check if error is fixed

### DIFF
--- a/src/components/BackupRestoreModal.tsx
+++ b/src/components/BackupRestoreModal.tsx
@@ -1,15 +1,14 @@
 import React from "react";
+import { Transaction } from "dexie";
 import { useRecoilValue, useSetRecoilState } from "recoil";
 
 import { db } from "@models";
+import ZModal from "@src/common/ZModal";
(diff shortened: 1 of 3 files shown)


tijlleenders:
@vinaybadgujar102 Nope, still same error.  
I also think it is probably related to the DB schema change. Maybe @Tushar-4781 can take a look? He's encountered/handled this kind of upgrade issue many times.

tijlleenders (approved):

Issue #1917, opened 2024-05-18
Catch scheduler error


tijlleenders:
**Is your feature request related to a problem? Please describe.**
It's very annoying when you can't use Goals/Feelings when there is an error calculating the schedule. Now whenever there is an error on mobile the whole app becomes unusable with only a white screen.

**Describe the solution you'd like**
1. Catch and show the scheduler error in a modal.
2. Don't update the cached version - so show the previous (working) output of the scheduler after closing the modal.

**Describe alternatives you've considered**
None.

**Additional context**
None.

tijlleenders:
@vinaybadgujar102 Just only show the error message in a modal and don't update the cache / show the latest working cached schedule.

vinaybadgujar102:
@tijlleenders What should be the error message shown in the modal?
-  Error Title?
- Description?

Should the error modal be shown only the first time the scheduler breaks, or every time the user opens the app?
What should the error modal should look like?

tijlleenders:
@vinaybadgujar102 I expect to see the same output as in the console, if that's possible. How it looks is not very important. This is something that should rarely happen.

Show the error every time it happens.

tijlleenders:
@vinaybadgujar102 The only thing I see is this:
I'd expect a more detailed message, like in the console.

![Screenshot_20240604-190419.png](https://github.com/tijlleenders/ZinZen/assets/44887755/f83971fd-8fa3-4969-b10d-95d6c5363f49)



Also, it only appears once, on app start.

tijlleenders:
@vinaybadgujar102 Is the following possible?
![image](https://github.com/tijlleenders/ZinZen/assets/44887755/f97e6baf-e6d6-410b-bec2-60307a145c9a)

vinaybadgujar102:
@tijlleenders, I couldn't find a way to capture the errors thrown by the scheduler module.

PR #1922, opened 2024-05-23
show error message in a modal when scheduler breaks

vinaybadgujar102:
#1917

### DIFF
--- a/pkg/scheduler.d.ts
+++ b/pkg/scheduler.d.ts
@@ -1,11 +1,14 @@
 /* tslint:disable */
 /* eslint-disable */
+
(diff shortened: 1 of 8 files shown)


tijlleenders:
@vinaybadgujar102 Any idea how to break the scheduler, so I can test this?

vinaybadgujar102:
@tijlleenders I can push a commit temporarily which breaks the scheduler.

tijlleenders (approved):
Awesome!

tijlleenders:
@vinaybadgujar102 Can you remove the commit so we can merge this?

tijlleenders (approved):

PR #1923, opened 2024-05-23
update timeBudget field in goalItem to be budget object or undefined

### LINKED ISSUE #1760: Handle Possibly Null Values in timeBudget Property of GoalItem Interface
**Describe the bug**
The GoalItem interface includes a timeBudget property that consists of perDay and perWeek fields, both of type string | null. We need to handle possibly null values in the timeBudget property to avoid potential runtime errors.

**Proposed Changes:**
Don't save "timeBudget" property in "goalCollection" db if its values are undefined/null

vinaybadgujar102:
resolves #1760

### DIFF
--- a/src/models/GoalItem.ts
+++ b/src/models/GoalItem.ts
@@ -28,10 +28,12 @@ export interface GoalItem {
   link: string | null;
   participants: IParticipant[];
   rootGoalId: string;
-  timeBudget: {
-    perDay: string | null;
-    perWeek: string | null;
-  } | null;
+  timeBudget:
+    | {
+        perDay: string;
+        perWeek: string;
+      }
+    | undefined;
   typeOfGoal: "myGoal" | "shared";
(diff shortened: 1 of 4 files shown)


tijlleenders:
@vinaybadgujar102 I've asked @Tushar-4781 to review this.

src/models/GoalItem.ts
```diff
@@ -28,10 +28,12 @@ export interface GoalItem {
   link: string | null;
   participants: IParticipant[];
   rootGoalId: string;
-  timeBudget: {
-    perDay: string | null;
-    perWeek: string | null;
-  } | null;
+  timeBudget:
```
Tushar-4781, line 31: you can write this instead
timeBudget?:
{
        perDay: string;
        perWeek: string;
      }

Tushar-4781 (approved):

PR #1928, opened 2024-05-27
implement feature to move goal

### LINKED ISSUE #1918: Move/Paste Goal/Budget in hierarchy
It's very annoying when you need move Goals/Budgets atound but can't. Especially with a few (layers) of subGoals it's not trivial to just delete/recreate.

1. Add a 'Move' option to the edit modal
2. When 'Move' is pressed, show a toast 'Navigate to destination place and confirm via +'
3. When in a different place of the hierarchy, show an extra + option 'Move here'.

Not using the 'normal' navigation to go to the destination place. Less intuitive.

**Additional context**
See info in comments before starting this.

vinaybadgujar102:
resolves #1918


vinaybadgujar102:
@tijlleenders there is move button in edit modal. clicking it activates move feature
image

vinaybadgujar102:
> @vinaybadgujar102 Ah ok. I expect it to be in the first modal so not to crowd the edit modal.

@tijlleenders, should I move the move option to first modal (goal action modal)?

tijlleenders:
@vinaybadgujar102 Yes please.

tijlleenders (changes requested):
@vinaybadgujar102 I think I didn't explain the feature well enough. 

The 'move here' button should not be visible everywhere. There should just be another option when you hit the plus button, just like options 'Goal' and 'Budget' there would be an option 'Paste'.

What happens if you move Goals that are shared with a partner?

vinaybadgujar102:
@tijlleenders 
> What happens if you move Goals that are shared with a partner?

Can this be handled in another PR? This PR already has a lot of changes, and I currently have no understanding of the code related to sharing/collaboration. I need to understand it, which may take some time.

tijlleenders:
@vinaybadgujar102 Sorry, I can't merge this if it breaks the sharing.  

@Tushar-4781 Can you please advise if this will impact sharing - since you already know the sharing code?  
If so - we might need to rethink the structure of the data and/or the method of sharing.

tijlleenders:
It shouldn't affect anything. 
- You should be able to move your goals around without it having any effect on sharing. Updates should still work.
- It should remain clear what is shared and what is not.

Tushar-4781:
@vinaybadgujar102 the moving feature is not that straight forward because there is a traversal mechanism to collect the ancestors of goal so moving a subtree somewhere else will create a problem in the traversal there is nothing to review / fix its a feature that has to be implemented at view layer ( which is something don't exist and we need to create it ) this might bring some changes in how we fetch and list goals on the UI. i m working on it so for now i would recommend to avoid working on anything big. although pick any other minor issues from backlog to work on.

tijlleenders:
@Tushar-4781 All major work has been paused for 30 days now.  
Is there any ETA on this PR landing?

Tushar-4781:
@tijlleenders the decoupling will take time there is no way to merge this PR. U may resume the existing work

tijlleenders:
@Tushar-4781 is implementing this in `develop` branch so this PR can be closed. It is a little more complex than we thought :)  
Thanks for the effort @vinaybadgujar102 !

PR #1933, opened 2024-06-08
regularly show scheduler error modal and handle cached schedule storage on scheduler crash 

### LINKED ISSUE #1917: Catch scheduler error

vinaybadgujar102:
resolves #1917

### DIFF
--- a/pkg/scheduler.d.ts
+++ b/pkg/scheduler.d.ts
@@ -8,7 +8,7 @@ import { ISchedulerInput, ISchedulerOutput } from "@src/Interfaces/IScheduler";
  * @param {ISchedulerInput} input
  * @returns {ISchedulerOutput}
  */
-export function schedule(input: ISchedulerInput): ISchedulerOutput;
+export function schedule(input: ISchedulerInput): ISchedulerOutput | undefined;
 
 interface Input {
   startDate: string;
--- a/src/components/MyTimeComponents/SchedulerErrorModal.tsx
+++ b/src/components/MyTimeComponents/SchedulerErrorModal.tsx
@@ -1,23 +1,23 @@
+/* eslint-disable react/no-array-index-key */
 import React from "react";
 import "../index.scss";
 import ZModal from "@src/common/ZModal";
 import { useRecoilState } from "recoil";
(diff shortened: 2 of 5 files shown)


tijlleenders:
@vinaybadgujar102 No, this is not showing the error still. Thanks for trying. 
Let's just park the issue and move on.

vinaybadgujar102:
@tijlleenders This PR addresses the issue where the error is only logged once when the scheduler fails. Previously, when the scheduler errored, the cached schedule was stored as undefined in the database. This undefined response caused further issues downstream. Additionally, since the input remained the same, the scheduler wouldn't run again, leaving the error unlogged. This PR resolves these issues.

tijlleenders (approved):
Ah OK, well that indeed works.
@Tushar-4781  Good to merge?

Tushar-4781:
@vinaybadgujar102 could you please share a test or step to verify the work and o/p of this mechanism ?

vinaybadgujar102:
@Tushar-4781, Steps to break the scheduler:
1. Reschedule any task to 7 days.
2. Change the system's date to 2 days ahead.
3. This breaks the scheduler and shows an error as a modal on the My Time page.

>This PR addresses the issue where the scheduler error is only logged once when the scheduler fails. Previously, when the scheduler errored, the cached schedule was stored as undefined in the database. This undefined response caused further issues downstream. Additionally, since the input remained the same, the scheduler wouldn't run again, leaving the error unlogged. This PR resolves these issues.

tijlleenders:
I tested manually and can confirm a breaking scheduler shows an error message every time you visit the My time page, instead of once on app startup only.  

Don't think it's worth the effort to make an automated test for this - as it seems tricky to realize...  

When this PR merges - we'll need a separate issue on the backlog (low prio) for showing better error messages (like in the console).

PR #1939, opened 2024-06-23
Revamp: Journal Page

Tushar-4781:
- React 17 -> 18
- Added: React Query package 
- pruned unwanted scss and created short custom classes ( later can be migrated to tailwind )
- infinite scrolling in journal page
- feelings api(s) simplified


tijlleenders (approved):
Awesome!!

PR #1951, opened 2024-07-12
Nick/1302/version number on feedback page

Nicholas-Kipkoech:


tijlleenders:
@Nicholas-Kipkoech  Please fix the linta before asking  for a review.

Nicholas-Kipkoech:
@tijlleenders  Fixing it.

tijlleenders:
@Nicholas-Kipkoech  Thanks that looks real good!

tijlleenders:
On the content however, the hash that is calculated is not really useful. It needs to give us a way to relate to a particular build or release of the app - so that we can check the bug or verify we are testing the right version.

image

I was hoping to use the hash shown in the Vite build files (see above screenshot) - but anything that refers us to an exact version of the code will do.

tijlleenders (changes requested):
Is above comment enough for you to work with?

Nicholas-Kipkoech:
@tijlleenders Yes it is now clear let me make the changes requested.

Nicholas-Kipkoech:
After some change this is the hash which is going to be generated
 ` version.json created: { buildHash: 'b66a66a2', buildDate: '2024-07-13T10:02:12.663Z' }`

Nicholas-Kipkoech:
@tijlleenders Changes addressed.

tijlleenders:
Perfect!

Is it possible to include the second hash, for the `index.js` as well?
image

Nicholas-Kipkoech:
@tijlleenders Yes it is possible. Can we join the two hashes when rendering on the page? Maybe separating by dash?

tijlleenders:
@Nicholas-Kipkoech Sure, separate them by a comma and a space (`, `).

tijlleenders:
@Nicholas-Kipkoech Can you resolve the merge conflicts?

tijlleenders (approved):

Tushar-4781 (approved):

tijlleenders:
@Tushar-4781 There were some extra dependencies added in this PR.  
Are these really necessary - and does it impact the bundle size?  
They are not in the dev-dependencies.

Nicholas-Kipkoech:
@tijlleenders We can remove webpack packages since we are not using it anywhere

tijlleenders:
@Nicholas-Kipkoech Please make a separate PR for remiving unused dependencies - although if we don't use it it should not impact the release bundle size. Stll, from 'clean code' perspective a good thing to do!

PR #1952, opened 2024-07-14
added dnd kit packages for draggable components

Nicholas-Kipkoech:


Nicholas-Kipkoech:
Works better for desktop. Currently working on improving the mobile devices its quite buggy.

Nicholas-Kipkoech:
@tijlleenders  I have used dnd-kit on this and the drag activation is so fast now.

tijlleenders (changes requested):
@Nicholas-Kipkoech That's great. It proves that it's possible. DnD was one option - not necessarily the best but if it works let's use it.

Did you try using it though?  
It's not really useable in my opinion without adding some improvements:  

- Only activate drag on the circle  
- Bring back the edit functionality when tapping on circle once  
- Only allow vertical movement of goals

Nicholas-Kipkoech:
@tijlleenders  This is noted.

tijlleenders:
PS Also think of remiving any previous drag&drop libraries.

tijlleenders (changes requested):
This is so nice @Nicholas-Kipkoech !

I do notice that the drag is still activated when touching outside the (box surrounding) the circle, like the goal titles. This makes it impossible to scroll a long list, or refresh the page.

Do you see this too?

Nicholas-Kipkoech:
@tijlleenders Yes, I have seen it. Let me try again and test.

Nicholas-Kipkoech:
@tijlleenders  I believe I have addressed the issues we were having.

tijlleenders (changes requested):
Yes, this is nice.  

Something else:  
When reordering, the reordering used to be persistent. When you reorder, then navigate to a subGoal and later come back, the reordering would still be there. In this PR, the ordering gets reset when you navigate away, then come back.    

Can you look at the previous code and re-implement persistent ordering in the same way?

Nicholas-Kipkoech:
@tijlleenders Noted.

tijlleenders (approved):
Nice work @Nicholas-Kipkoech !

This is a huge UX improvement.


@Tushar-4781 Can you please do a code review?

Tushar-4781 (approved):

PR #1953, opened 2024-07-17
822/strikethrough goal text when done hide done when revisiting

Nicholas-Kipkoech:


Nicholas-Kipkoech:
@tijlleenders Can we reduce the time it takes for a goal when marked done to get archived?

tijlleenders:
@Nicholas-Kipkoech You mean that instead of waiting to archive for me to navigate away and come back, we wait XX seconds? That could indeed be an option. But what is the advantage? And what happens if you navigate away when the goal is not archived yet?

Nicholas-Kipkoech:
@tijlleenders  You can still navigate to any other page and the goal will be archived because it is a background process it wont affect other functionalities. The default wait now is 3 seconds we can reduce to 2 seconds maybe?

Nicholas-Kipkoech:
@tijlleenders  You can still navigate to any other page and the goal will be archived because it is a background process it wont affect other functionalities. The default wait now is 3 seconds we can reduce to 2 seconds maybe?

tijlleenders (changes requested):
Ok, this works. Just:  
- from a user perspective, the strikethrough should coincide with the completion sound, not with the move to archive.  
- increase the wait to 10 seconds

Nicholas-Kipkoech:
@tijlleenders noted!

tijlleenders (approved):
This is good ... for now.  

I realized that with the timeout it might getba little unexpected as goals can change / shift positions unexpectedly.  

Like I said ok for now - but the better option is to have them disappear only after you navigate away.

@Tushar-4781 Please review the code.

Nicholas-Kipkoech:
@tijlleenders For them to disappear when you navigate away is to reduce the delay to 2 seconds because already I have synced the sound with line stroke

tijlleenders:
@Nicholas-Kipkoech That is not what I meant. I meant that ideally, the goals stay crossed out, forever, until you navigate away.

tijlleenders:
@Tushar-4781 *bump*

Tushar-4781 (approved):

PR #1998, opened 2024-07-28
creating-sub-goalbudget-causes-error

Nicholas-Kipkoech:


tijlleenders:
@Nicholas-Kipkoech Be careful to respect the differences in Goal approach between main and develop branch. Some things are fundamentally different in develop, which means the way we used to do it in main will be abandoned.  

If in doubt, ask @Tushar-4781

Nicholas-Kipkoech:
@tijlleenders If this is working in main, then we don't have to do this on develop since we will be repeating ourselves. Maybe we can check later when the current develop has been merged with the main branch.

tijlleenders:
@Nicholas-Kipkoech 
please leave this issue for @Tushar-4781

Issue #1987, opened 2024-07-26
Creating sub goal/budget causes error


vinaybadgujar102:
**Describe the bug**
After creating sub goal/budget, app breaks.

**Screencast**
[Screencast from 26-07-24 08:31:05 PM IST.webm](https://github.com/user-attachments/assets/3a08d3b8-ce06-4f73-885d-193e6e67b3ea)

**Console Error**

![Screenshot from 2024-07-26 20-34-50](https://github.com/user-attachments/assets/a5efc238-3f07-48c9-add1-8db0a6fdffb2)

tijlleenders:
@vinaybadgujar102 Can you be a bit more specific by adding a screencast and console errors?  
Or are you now busy implementing the fix already?

vinaybadgujar102:
> @vinaybadgujar102 Can you be a bit more specific by adding a screencast and console errors? Or are you now busy implementing the fix already?

updated the description

Nicholas-Kipkoech:
@vinaybadgujar102 I noticed this issue today, was this working on main branch?

tijlleenders:
@Nicholas-Kipkoech You can test this yourself, locally or on ZinZen.vercel.app  ;)

Issue #1994, opened 2024-07-27
Remove lowest level breadcrumb


tijlleenders:
**Is your feature request related to a problem? Please describe.**
It's very annoying when there is redundant info on screen. It crowds the screen and makes it busier than necessary.

**Describe the solution you'd like**
Remove the lowest level of the breadcrumb like so:  
![IMG_20240727_115731.jpg](https://github.com/user-attachments/assets/d6dd321a-790a-4b22-88c9-342eae447e0e)

**Describe alternatives you've considered**
None.

**Additional context**
None.

PR #2002, opened 2024-07-29
fix: fixed Creating sub goal/budget causes error

Nicholas-Kipkoech:


tijlleenders (changes requested):
Please fix breadcrumb so I can test this.

Nicholas-Kipkoech:
That would be great so that we can do it all at once

Nicholas-Kipkoech:
@tijlleenders You can close the breadcrumb issue since it is fine now.

Nicholas-Kipkoech:
Do we need to remove the lowest level breadcrumb? so that we can only remain with parent?

tijlleenders:
@Nicholas-Kipkoech Please mention which issue(s) this PR closes.

Unless otherwise required, keep with one issue per PR.

tijlleenders (changes requested):
See comment above

Nicholas-Kipkoech:
This PR will resolve #1987 and #1994

tijlleenders:
> This PR will resolve #1987 and #1994

Please always add this to the PR description, like so:
```
Resolves #1987 
Resolves #1994 
```

I doubt #1994 should be included, it is a modification on the working breadcrumb. We currently have a breadcrumb that is not functioning like in main.   
Maybe you should make an new issue for the bug and reference that, if that bug doesn't exist already.

Nicholas-Kipkoech:
@tijlleenders This goal will only resolve #1987 and #1994 will be an independent issue on its own.

tijlleenders:
> @tijlleenders This goal will only resolve #1987 and #1994 will be an independent issue on its own.

Again: please add it to the PR description.

tijlleenders (approved):

tijlleenders:
@Tushar-4781 Please review the code.

Tushar-4781 (approved):

PR #2007, opened 2024-07-31
skip implementation

### LINKED ISSUE #1866: Implement state for skipped Activities
It's very confusing when you skip an Task on the Calendar - and 
- it then get crossed out, as if it was actually done
- it resets on a refresh / navigate away/come back

- Remember what is skipped for the current day ~~week so it doesn't reset on a new scheduler calculation~~
- ~~reduce the hours of the corresponding budget min/max per week, or goals sent to the scheduler~~
- ~~provide a global not-on to the scheduler to keep that slot completely empty~~
- reset the skipped hours at the end of the **day** ~~week~~  

**Added after discussion in PR:**  
- Don't change the information sent to the scheduler (a new calculation is not even required) - but mask the skipped Task visually in the frontend (show the slot as free even though scheduler has the Task there).
- Remove 'Skip' and 'Reschedule' as options on Tasks.
- Replace them with a 'Not now' option that opens a 'Not now' modal.
- The 'Not now' modal has the same options as current 'Postpone' modal, but an added option 'Skip' when the Task is the result of a Budget. When the Task is from a simple Goal, there is no such option.

None.

**Additional context**
#1400 provides a 'Reschedule' button for Tasks, next to the Skip.
We could simplify the submenu for tasks by - merging 'Skip' and 'Reschedule' into a generic 'Not now' - wand adding a skip option to the existing reschedule (renamed to 'Not now') options popup.

vinaybadgujar102:
resolves #1866 

> reduce the hours of the corresponding budget min/max per week, or goals sent to the scheduler

@Tushar-4781 Based on the above issue description, I implemented logic to reduce budget min/max weeks. However, as we skip entire task for day of a particular goal, we also have to remove that day from being passed to the scheduler. I am currently stuck on how to remove that particular day from onDay array.

Additionally, could you please explain the purpose of the forgotToday field in TaskItem?


tijlleenders:
> @Tushar-4781 Based on the above issue description, I implemented logic to reduce budget min/max weeks. However, as we skip entire task for day of a particular goal, we also have to remove that day from being passed to the scheduler. I am currently stuck on how to remove that particular day from onDay array.

@vinaybadgujar102 You don't have to change the onDay array - just add a notOn for the specific day. This, in combination with the reduced min/max per week will have the desired effect.

vinaybadgujar102:
@tijlleenders, adding notOn of entire day for skipped task is not working. You can check it.

tijlleenders:
Please add the API call to the scheduler. I'll fix and add a new scheduler version to this PR tomorrow.

tijlleenders:
Waiting to see the API call but block should only be for the start-enddatetime of the skipped task, not the entire day. Otherwise nothing at all will be scheduled that day...

vinaybadgujar102:
> Please add the API call to the scheduler. I'll fix and add a new scheduler version to this PR tomorrow.

@tijlleenders, API should be like this?

```
export interface ISchedulerInput {
  startDate: string;
  endDate: string;
  goals: ISchedulerInputGoal[];
  tasksCompletedToday: TCompletedTaskTiming[];
  tasksForgotToday: TCompletedTaskTiming[];
  globalNotOn: [
     { goalId: 
          [ 
             {start, end} 
          ] 
      }
]}

```

tijlleenders:
Yeah, but an example of what is actually sent (a testcase) would be nice.

vinaybadgujar102:
@tijlleenders 
```
{
  "startDate": "2024-08-01T00:00:00",
  "endDate": "2024-08-08T00:00:00",
  "goals": [
    {
      "id": "208360a4-fb71-44c6-b22d-4053100b389e",
      "title": "Work 💪🏽",
      "filters": {
        "afterTime": 6,
        "beforeTime": 18,
        "onDays": [
          "mon",
          "tue",
          "wed",
          "thu",
          "fri"
        ]
      },
      "createdAt": "2024-08-01T13:27:55.424Z",
      "notOn": [
        {
          "start": "2024-08-01T06:00:00",
          "end": "2024-08-01T14:00:00"
        }
      ],
      "budget": {
        "minPerWeek": 32,
        "maxPerWeek": 32,
        "minPerDay": 6,
        "maxPerDay": 10
      }
    }
  ],
  "tasksCompletedToday": [],
  "tasksForgotToday": [
    {
      "goalid": "208360a4-fb71-44c6-b22d-4053100b389e",
      "start": "2024-08-01T06:00:00",
      "deadline": "2024-08-01T14:00:00"
    }
  ],
  "globalNotOn": [
    {
      "208360a4-fb71-44c6-b22d-4053100b389e": [
        {
          "start": "2024-08-01T06:00:00",
          "end": "2024-08-01T14:00:00"
        }
      ]
    }
  ]
}
```

tijlleenders:
@vinaybadgujar102 The globalNotOn should not reference any goals, it is global. Just an array of startdatetime,enddatetime tuples.

vinaybadgujar102:
@tijlleenders, it still doesnt work. Try skipping "Me time"

tijlleenders:
Did you read the error message and check if you did the sum correctly?  
image

tijlleenders:
Min per day should be 0, if we're skipping it on a certain day.  

That or removing the OnDay ... 
... which is probably the better approach, compared to reducing min per day, at least while we're only calculating one week in advance.

{"id":"5f728e9e-0dec-4a8b-a20f-40e7313db9b2",
"title":"Me time 🧘🏽😌",
"filters":{"afterTime":5,"beforeTime":23,
"onDays":["mon","tue","wed","thu","fri","sat","sun"]},
"createdAt":"2024-08-03T07:08:18.227Z",
"notOn":[{"start":"2024-08-03T05:00:00","end":"2024-08-03T06:00:00"}],
"budget":{"minPerWeek":6,"maxPerWeek":6,
**"minPerDay":1**,"maxPerDay":1}}

vinaybadgujar102:
> . However, as we skip entire task for day of a particular goal, we also have to remove that day from being passed to the scheduler. I am currently stuck on how to remove that particular day from onDay array.

@tijlleenders, we need to remove an entire day from onDay array. 
The 'My time' goal is set for one hour daily over 7 days, weekly min-max of 7 hours. If we skip a day, we reduce the weekly min-max by 1 hour and then pass this updated value to the scheduler. To ensure the scheduler functions correctly, we must also remove the corresponding day from the onDay array which will also reduce weekly hours to 6 (1 x 6 days).

I have already tried removing the day from the onDay array, and it worked. However, I couldn't develop a logic to remove the specific skipped day of the week.

tijlleenders:
I'm currently on the go ... so difficult to concentrate on this.  

I feel the scheduler might need more adjustments. Instead of doing the adjustment of the budget in the frontend, the better approach is just to pass a `notOn` for the whole day you want to skip the task for, and the scheduler should adjust the corresponding day and week budget.  

Let me see if I can implement that this weekend.

tijlleenders:
That way:  
- we keep complexity about scheduling out of the frontend (just declarative syntax)
- it will also work if we extend the requested calendar to more than a week

tijlleenders:
@vinaybadgujar102 I sat down to think about this ... and actually the solution is easier than I thought. Sorry, I did not think this through enough before putting it on the backlog - or maybe this is the feedback I needed to iterate...

So, basically we need to (re)think about the use cases for 'skip' - and what 'skip' means in context:
- For Goals it doesn't make sense - you can only postpone the work. Skipping is an alias for postpone when we 'skip' 'Writing a report 1h'. To avoid confusion, 'skip' should not be an option in the UI for Goals.
- It only makes sense for a generic schedule/budget. You can 'skip' 1h of "Work 40h/week" or 'skip' 1h of "Exercise 3h/week, 0-1h/day".
  
The `globalNotOn` is not required.  
We'll keep the code/implementation for `globalNotOn` as I'm sure there's a use case for telling the scheduler to keep certain datetime ranges completely free.

In conclusion we need to implement the suggestion in 'additional context' of the issue:  
- Remove 'Skip' and 'Reschedule' as options on Tasks.  
- Replace them with a 'Not now' option that opens a 'Not now' modal.
- The 'Not now' modal has the same options as current 'Postpone' modal, but an added option 'Skip' when the Task is the result of a Budget. When the Task is from a simple Goal, there is no such option.  

I adapted the issue description, let me know if there are questions - we can voice.

vinaybadgujar102:
Closing this as issue description is changed. Will resolve in new PR.

Issue #1989, opened 2024-07-26
Completing starter goals gives weird outcome


tijlleenders:
**Describe the bug**
Something goes wrong when completing multiple goals.

**To Reproduce**
[Screencast from 2024-07-26 16-58-25.webm](https://github.com/user-attachments/assets/3bfa1e88-7057-4104-b1bb-72c02dfe0d00)
No console errors.

**Expected behavior**
Completing multiple goals should archive all of them.
They should only be visible in the Trash of the Goal they were in.

**Additional context**
None.

Nicholas-Kipkoech:
@tijlleenders Have tested this and it seems to be working fine.

Nicholas-Kipkoech:
Seems the issue is in the subgoals.

tijlleenders:
@Nicholas-Kipkoech Maybe better to skip this until @vinaybadgujar102 finishes #822

vinaybadgujar102:
@tijlleenders, I think has been resolved

tijlleenders:
Yes, closing this as I can't reproduce this in `develop` branch.

PR #2010, opened 2024-08-03
strikethrough done goals

### LINKED ISSUE #822: Strikethrough goal text when done - hide done when revisiting
Strikethrough when marking it done (just like on MyTime page) - so don't hide it.

After navigating away and coming back - hide it in the archived area at the bottom.  
It would be nice if the strikethrough has the color of the goal.

**Additional context**   
This issue was previously implemented, but not in the way as specified in the issue.  
The crossed out goal now remains for a number of seconds, then suddenly disappears. This is surprising, confusing and also makes the position of the Goals suddenly change, causing you frustration when you want to tap some goal, and right before you hit it, it shifts to some other goal.

vinaybadgujar102:
resolves #822


tijlleenders (changes requested):
This is not working as expected.

When I complete multiple goals, suddenly a subset of them appears in Done, even though I haven't navigated away yet:
image


When I do navigate and come back, I expect them all to be in 'Done' - without any crossed out. Instead I see this:  
image

Tushar-4781 (changes requested):
Please change the implementation to jquery style. the feature is just a temporary change on UI. we don't need to use global state for this.

vinaybadgujar102:
> Please change the implementation to jquery style. the feature is just a temporary change on UI. we don't need to use global state for this.

Please elaborate. I am not familiar with jQuery.

Tushar-4781:
> > Please change the implementation to jquery style. the feature is just a temporary change on UI. we don't need to use global state for this.
> 
> @Tushar-4781, Please elaborate. I am not familiar with jQuery.

@vinaybadgujar102 directly manipulate the DOM instead of doing it via react state.

vinaybadgujar102:
> This is not working as expected.
> 
> When I complete multiple goals, suddenly a subset of them appears in Done, even though I haven't navigated away yet

@tijlleenders, the subgoals location state has some bugs. It is related to location history I guess. 
https://github.com/tijlleenders/ZinZen/issues/1989#issuecomment-2264837583 It should be handled in a separate issue.

vinaybadgujar102:
@Tushar-4781 changed the approach as suggested

Tushar-4781 (changes requested):
@vinaybadgujar102 can you try directly accessing the title component by using the goal id and then using the class name goal-title. we don't the ref or prop either

tijlleenders (changes requested):
@vinaybadgujar102 Please keep this as draft until it is ready to be testd again. Currently not working correctly when completing multiple goals.

tijlleenders (changes requested):
Works as expected.
Can we make the strikethough line a bit thicker?

I can hardly see breakfast is crossed out here:

tijlleenders (changes requested):
You can double the thickness.

tijlleenders (approved):
Great, thanks!

tijlleenders:
@Tushar-4781 Changes OK?

Tushar-4781 (approved):

PR #2013, opened 2024-08-04
introduce "Not now" task option and move skip to not now modal

### LINKED ISSUE #2014: move skip task option in new "Now now" modal
As mentioned in issue #1866, remove the "Skip" and "Reschedule" task options and add a new "Not Now" option. Move the "Skip" option to the "Not Now" modal, and only display it if the task is related to a budget.

@tijlleenders, I've created this separate issue to break down the skip implementation into more manageable smaller tasks.

vinaybadgujar102:
resolves #2014 

P.S.: The Goal/Task Action modal code needs to be refactored. The current implementation makes it difficult to maintain and apply styles. Will create a new issue for it.


tijlleenders (approved):
Looks good.

vinaybadgujar102:
@Tushar-4781 ready for review

src/components/MyTimeComponents/NotNow/NotNowModal.tsx
```diff
@@ -47,9 +69,15 @@ const Reschedule = () => {
...
+            <ActionDiv label="Skip" icon="Clock" />
+          </div>
+        )}
       </div>
     </ZModal>
   );
 };
 
-export default Reschedule;
+export default NowNowModal;
```
Tushar-4781, line 83: what's this ?

Tushar-4781 (changes requested):

Tushar-4781 (approved):

tijlleenders (approved):
I'm not sure about the wording in the 'Not now' modal. I'm inclined to use 'postpone' but if you 'postpone' for 1h it could also be scheduled even earlier in the day - as you on have indicated the current slot ia not convenient...

Anyway. This is mergeable. We'll use this for a while and see.

PR #2015, opened 2024-08-05
subgoalhistory bug

### LINKED ISSUE #2016: navigating to different page from subgoal doesnt clear subGoalHistory location state
**Describe the bug**
When the user is on the sub-goal page and navigates away, for instance, to the "My Time" page, the sub-goal state is not cleared. Consequently, if the user returns to the sub-goal page, the same state is pushed to the sub-goal state array, causing bugs including the one mentioned here (https://github.com/tijlleenders/ZinZen/pull/2010#pullrequestreview-2217007450). 

**Screenshots**
[Screencast from 05-08-24 01:02:10 PM IST.webm](https://github.com/user-attachments/assets/9ccfcbcc-aa20-475b-a8d7-bda18a3b47d9)

vinaybadgujar102:
resolves #2016

### DIFF
--- a/src/components/BottomNavbar/BottomNavbar.tsx
+++ b/src/components/BottomNavbar/BottomNavbar.tsx
@@ -52,8 +52,8 @@ const BottomNavbar = ({ title }: { title: string }) => {
       const newLocationState = { ...location.state, from: currentPage, displayFocus: false };
       if (to === "MyTime") {
         if (currentPage !== "") navigate("/", { state: newLocationState });
-      } else if (to === "MyGoals") {
-        if (currentPage !== "MyGoals") {
+      } else if (to === "goals") {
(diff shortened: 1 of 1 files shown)


tijlleenders (changes requested):
It is hard to see what this fixes as the behavior in the PR is not correct. When I switch to MyTime from a subgoal and return, I expect to still be in the subGoal. Instead I'm returned to root level.

vinaybadgujar102:
> When I switch to MyTime from a subgoal and return, I expect to still be in the subGoal. Instead I'm returned to root level.

@tijlleenders. This behaviour was introduced in develop branch and I thought this was what was expected. I should have checked the main branch.

You can review now.

tijlleenders (approved):

Tushar-4781 (approved):

PR #2019, opened 2024-08-11
add goal from my time page

### LINKED ISSUE #1958: Add a Goal from My Time page
It's very annoying when you open the app to quickly add a Goal, and first have to switch to My Time page to add a Goal.

- Add a '+' button on the My Time page, just like on My Goals.
- When adding a Goal via My Time page, don't leave the My Time page - just show the add Goal modal
- When saving, regardless of where you are in the Goal hierarchy on the My Goals page, since you added via My Time in a hurry, there is no reasonable place to put it, other than the root level.  
This way you can add quickly, even on My Time, and sort out where you want it in you hierarchy later.

None.

**Additional context**
#1957 Needs to be done first.

vinaybadgujar102:
resolves #1958

### DIFF
--- a/src/components/index.scss
+++ b/src/components/index.scss
@@ -33,29 +33,43 @@
   border-radius: 30px;
   position: absolute;
   border: 1px solid var(--icon-grad-1);
-  color: white;
   font-size: 18px;
   justify-content: space-between;
   align-items: center;
   background-color: var(--secondary-background);
   padding: 5px;
   width: 150px;
-}
+  right: 35px;
 
-.goal-btn-circle {
-  border-radius: 50%;
-  border: none;
-  background: var(--selection-color);
-  box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.25);
-  width: 56px;
-  height: 56px;
-  color: white;
-  font-weight: 600;
-  font-size: 0.875em;
-  display: flex;
-  justify-content: center;
-  align-items: center;
-  cursor: pointer;
+  &.disabled {
+    opacity: 0.25;
+    pointer-events: none;
+  }
+
+  .button-text {
+    padding-left: 5px;
+  }
+
+  .goal-btn-circle {
+    border-radius: 50%;
+    border: none;
+    background: var(--selection-color);
+    box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.25);
+    width: 56px;
+    height: 56px;
+    color: white;
+    font-weight: 600;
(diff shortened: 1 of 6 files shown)


tijlleenders (changes requested):
Yeah this is good.

Do you also see the modal going 'off screen' at the top? It would be nice if we can force the modal to stick to the virtual keyboard. This is not the first time I see this so might be good to make a separate issue for that (although I think there was one at some point so it might be a regression).  

Apart from checking the issue history for that  
and  fixing it (or making a separate issue for it),  
- what I realize is that a sound confirmation is not enough here. I'd like something visual, like a toast message with "[first max 15 characters of goal title] created!" message, as part of this PR.

tijlleenders:
Modal going offscreen:

vinaybadgujar102:
@tijlleenders I will make a new issue for the modal offscreen problem. 
I have added the toast once new goal is added. You can review it.

tijlleenders (approved):
This works, thanks.

src/components/index.scss
```diff
@@ -33,29 +33,43 @@
...
+    border-radius: 50%;
+    border: none;
+    background: var(--selection-color);
+    box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.25);
+    width: 56px;
+    height: 56px;
+    color: white;
+    font-weight: 600;
+    font-size: 0.875em;
+    display: flex;
```
Tushar-4781, line 63: you can use "place-middle" in class name. check in short.scss

src/components/index.scss
```diff
@@ -33,29 +33,43 @@
...
+
+  .goal-btn-circle {
+    border-radius: 50%;
+    border: none;
+    background: var(--selection-color);
+    box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.25);
+    width: 56px;
+    height: 56px;
+    color: white;
+    font-weight: 600;
```
Tushar-4781, line 61: you can use fw-600 in class name

src/components/index.scss
```diff
@@ -33,29 +33,43 @@
...
+    font-size: 0.875em;
```
Tushar-4781, line 62: pick a font class present from short.scss

Tushar-4781 (changes requested):

src/components/index.scss
vinaybadgujar102: removed font-size as it was not needed

Tushar-4781 (approved):

PR #2022, opened 2024-08-14
navigate goals with arrow keys

### LINKED ISSUE #1985: Navigate Goals with arrow keys
It's very annoying when you are used to only using a keyboard and can´t.

On the MyGoals page, when pressing up/down navigate through the goals. Show which Goal you are on with a small vertical line on the left hand side of the circle/icon, in the same color as the Goal/Budget/Video/Snippet.
Pressing left/right will navigate out/into a Goal.

Highlight the whole goal. This is a lot (too much) visual change. If it can be done in a subtle way, why make fireworks?

**Additional context**
None.

vinaybadgujar102:
resolves #1985


tijlleenders (changes requested):
Excellent!  
The one thing missing is that when you navigate back up, the selector should also be visible, so that you can go left, then right immediately to return to where you were, instead of having to press down first.

vinaybadgujar102:
> Excellent!
> The one thing missing is that when you navigate back up, the selector should also be visible, so that you can go left, then right immediately to return to where you were, instead of having to press down first.

@tijleenders, I implemented the requested change. However, I encountered an issue that I couldn't resolve. When a user navigates to another page, such as the Journal Page, and then returns to the Goals Page, the previously focused goal remains focused. The expected behavior should be that the focus resets. I'm not sure if this is the ideal approach, but it's what I believe should happen.

tijlleenders (changes requested):
When I navigate to 'Walk' and press down, I get this error: "Error: Minified React error #185; visit https://reactjs.org/docs/error-decoder.html?invariant=185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings."

tijlleenders:
> > Excellent!
> > The one thing missing is that when you navigate back up, the selector should also be visible, so that you can go left, then right immediately to return to where you were, instead of having to press down first.
> 
> @tijlleenders , I implemented the requested change. However, I encountered an issue that I couldn't resolve. When a user navigates to another page, such as the Journal Page, and then returns to the Goals Page, the previously focused goal remains focused. The expected behavior should be that the focus resets. I'm not sure if this is the ideal approach, but it's what I believe should happen.

I think it's fine to keep the focus.

tijlleenders (changes requested):
This looks good now.

One tiny mod: When adding a new goal I´d like the focus to be on the newly created goal, instead of sticking to the nth in the list.

vinaybadgujar102:
> One tiny mod: When adding a new goal I´d like the focus to be on the newly created goal, instead of sticking to the nth in the list.

We can address this in a new issue, as this PR is already somewhat complex.

tijlleenders:
No I think it should be fixed here, as the highlighted goal switch (it jumps to the previous goal) after adding a goal is unexpected and confusing, thus a bug.

tijlleenders (changes requested):
Excellent.  

The one thing I'd like to add to this PR is that the line doesn't move the goal circle and title to the right a bit. It's a small thing, but annoying (or at least not adding any value) when things jump a few pixels when moving around the highlighter.

tijlleenders (changes requested):
A lot better, but there is still a small budge in the Circle/Title:  
[Screencast from 2024-08-28 12-22-23.webm](https://github.com/user-attachments/assets/73a1aeb9-3518-4f1c-acc8-54330bed5d00)

Don't spend too much time, but see if you can eliminate it completely please.

vinaybadgujar102:
> A lot better, but there is still a small budge in the Circle/Title:  
> 
> [Screencast from 2024-08-28 12-22-23.webm](https://github.com/user-attachments/assets/73a1aeb9-3518-4f1c-acc8-54330bed5d00)
> 
> 
> 
> Don't spend too much time, but see if you can eliminate it completely please.

I tried with many different values, but slight misalignment remains.

tijlleenders (approved):

tijlleenders:
Close enough then, @Tushar-4781 agree on code?

src/hooks/useGoalSelection.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    const url = extractLinks(goal.title);
+    if (url) {
+      const finalUrl = url.startsWith("http://") || url.startsWith("https://") ? url : `https://${url}`;
+      window.open(finalUrl, "_blank");
+      return;
+    }
+    const newState: ILocationState = {
+      ...location.state,
+      activeGoalId: goal.id,
```
Tushar-4781, line 38: we don't use this anymore the active goal id will be fetched from params

src/hooks/useGoalSelection.ts
```diff
@@ -0,0 +1,136 @@
...
+      foucusedGoalIndex: focusedIndex,
```
Tushar-4781, line 39: use a query parameter for this focusing

src/hooks/useGoalSelection.ts
```diff
@@ -0,0 +1,136 @@
...
+      setFocusedIndex((prevIndex) => {
+        const newIndex = (prevIndex + 1) % goals.length;
+        scrollIntoView(newIndex);
+        return newIndex;
+      });
+    }
+  }, [downPress, goals.length, scrollIntoView]);
+
+  // handle up key press
+  useEffect(() => {
```
Tushar-4781, line 96: please club the code of all the useEffect(s) in a single useEffect. use if else to check which piece to execute

src/hooks/useGoalSelection.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  // handle left key press
+  useEffect(() => {
+    if (disableKeyboardNavigation) return;
+
+    if (leftPress) {
+      if (location.pathname === "/goals") {
+        return;
+      }
+      setFocusedIndex(location.state.foucusedGoalIndex === undefined ? -1 : location.state.foucusedGoalIndex);
```
Tushar-4781, line 126: create a function wrapped in useCallback that takes the index and handles the focusedIndex state and scrollIntoView functionality for you

Tushar-4781 (changes requested):

Tushar-4781 (approved):

PR #2026, opened 2024-08-20
show goal update confirmation

### LINKED ISSUE #1916: Add confirmation on Goal update
**Additional context**
This was previously implemented on `main` branch thanks to @Shubham-Praj , just has to be ported to `develop`.

It's very annoying when you don't get confirmation of your changes. After Goal update the edit modal just closes.

1. A very brief Toast message 'Goal updated!'
2. A very short (compared to add Goal) sound
Not implemented in `main` - but should be implemented now: if there are no changes, don't show the modal.

A confirmation popup. This breaks the flow.

**Additional context**
None.

vinaybadgujar102:
resolves #1916

### DIFF
--- a/src/hooks/useGoalActions.tsx
+++ b/src/hooks/useGoalActions.tsx
@@ -1,4 +1,4 @@
-import { getAllLevelGoalsOfId, unarchiveUserGoal, updateSharedStatusOfGoal } from "@src/api/GoalsAPI";
+import { getAllLevelGoalsOfId, getGoal, unarchiveUserGoal, updateSharedStatusOfGoal } from "@src/api/GoalsAPI";
 import { getSharedWMGoalById } from "@src/api/SharedWMAPI";
 import { restoreGoal } from "@src/api/TrashAPI";
 import { createGoal, deleteGoal, deleteSharedGoal, modifyGoal } from "@src/helpers/GoalController";
@@ -13,6 +13,7 @@ import { useSetRecoilState } from "recoil";
 import { shareGoalWithContact } from "@src/services/contact.service";
 import { addToSharingQueue } from "@src/api/ContactsAPI";
 import { ILocationState } from "@src/Interfaces";
+import { hashObject } from "@src/utils";
 
 const useGoalActions = () => {
   const { state }: { state: ILocationState } = useLocation();
@@ -59,6 +60,8 @@ const useGoalActions = () => {
   };
 
   const updateGoal = async (goal: GoalItem, hints: boolean) => {
+    let changes = false;
+
     if (isPartnerModeActive) {
       let rootGoal = goal;
       if (state.goalsHistory && state.goalsHistory.length > 0) {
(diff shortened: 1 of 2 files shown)


tijlleenders (changes requested):
This was probably implemented three times now, once in main, once by Nicholas, and then again here...  

Anyway, it looks good.  
Please check why the hints toggle is not being persisted. It is probably an unrelated issue. If so, please file a bug for `develop` and this can be merged.

vinaybadgujar102:
The hint toggle thing is unrelated, Created a new issue for it.

tijlleenders (approved):
Great!

Tushar-4781 (commented):
@vinaybadgujar102 Can you explain a bit what exactly happening in code ?

vinaybadgujar102:
@Tushar-4781, When the updateGoal function is called to modify a goal, we first generate hashes for both the original goal and the updated goal object passed to the function. We then compare these hashes to detect any differences. If differences are detected, we update the goal in the database and display a confirmation toast; otherwise, we simply return.

src/hooks/useGoalActions.tsx
```diff
@@ -67,9 +72,23 @@ const useGoalActions = () => {
       }
       suggestChanges(rootGoal, goal, subGoalsHistory.length);
     } else {
-      await modifyGoal(goal.id, goal, [...ancestors, goal.id], hints);
+      const originalGoal = await getGoal(goal.id);
+      if (originalGoal) {
```
Tushar-4781, line 76: club them up -> ( originalGoal && hashObject(originalGoal) !== hashObject(goal))

Tushar-4781 (changes requested):

src/hooks/useGoalActions.tsx
```diff
@@ -67,9 +74,21 @@ const useGoalActions = () => {
       }
       suggestChanges(rootGoal, goal, subGoalsHistory.length);
     } else {
-      await modifyGoal(goal.id, goal, [...ancestors, goal.id], hints);
+      const originalGoal = activeGoal;
```
Tushar-4781, line 77: no need to duplicate this goal.

src/hooks/useGoalActions.tsx
```diff
@@ -67,9 +74,21 @@ const useGoalActions = () => {
...
+      if (originalGoal && hashObject(originalGoal) !== hashObject(goal)) {
```
Tushar-4781, line 78: use activeGoal directly just add a comment that we are comparing the old and updated version.

Tushar-4781 (changes requested):

src/hooks/useGoalActions.tsx
```diff
@@ -59,17 +63,31 @@ const useGoalActions = () => {
...
       suggestChanges(rootGoal, goal, subGoalsHistory.length);
-    } else {
+    } else if (activeGoal && hashObject(activeGoal) !== hashObject(goal)) {
+      // Comparing hashes of the old (activeGoal) and updated (goal) versions to check if the goal has changed
+      isGoalChanged = true;
       await modifyGoal(goal.id, goal, [...ancestors, goal.id], hints);
     }
-    setLastAction("goalUpdated");
+
+    if (isGoalChanged) {
```
Tushar-4781, line 82: i think you can just move this entire logic in the above block right ?

Tushar-4781 (approved):

PR #2027, opened 2024-08-21
remember what is skipped for the current day

### LINKED ISSUE #1866: Implement state for skipped Activities

vinaybadgujar102:
resolves #1866

### DIFF
--- a/src/api/TasksAPI/index.ts
+++ b/src/api/TasksAPI/index.ts
@@ -31,7 +31,7 @@ export const getTaskByGoalId = async (goalId: string) => {
 
 export const getForgetHrsCount = (task: TaskItem) => {
   let yesterdaysCount = 0;
-  task.forgotToday.forEach((slot) => {
+  task.skippedToday.forEach((slot) => {
     const [start, end] = slot.split("-");
     yesterdaysCount += Number(end) - Number(start);
   });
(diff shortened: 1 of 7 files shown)


tijlleenders:
@vinaybadgujar102 I can check this out on Friday.

tijlleenders (approved):
Awesome!

Tushar-4781 (approved):

PR #2028, opened 2024-08-21
show toast when running scheduler first time

### LINKED ISSUE #1816: Show toast when running scheduler for first time.
It's very annoying when you see an empty my time page for x seconds when you open the app for the first time.

Show a toast when there is no schedule jn cache: "Automagical scheduling of your goals in progress..."

1. Start calculating async on the choose language page already. By the time you get to mytime the result is ready. Possible? More complicated?
2. Precache the default schedule with the app - skipping the first calculation alltogether.

**Additional context**
None.

vinaybadgujar102:
resolves #1816

### DIFF
--- a/src/pages/MyTimePage/MyTimePage.scss
+++ b/src/pages/MyTimePage/MyTimePage.scss
@@ -51,10 +51,5 @@
 }
 
 .scheduling-message {
-  text-align: center;
-  padding: 20px;
-  font-size: 1.275rem;
   height: 50%;
-  color: var(--text-color);
-  font-weight: bold;
 }
--- a/src/pages/MyTimePage/MyTimePage.scss
+++ b/src/pages/MyTimePage/MyTimePage.scss
@@ -49,3 +49,13 @@
   left: 4%;
   border-radius: 8px;
 }
+
+.scheduling-message {
+  text-align: center;
+  padding: 20px;
+  font-size: 1.275rem;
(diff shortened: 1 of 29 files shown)


tijlleenders (changes requested):
Upon review, the Toast approach is not the right one. It will show the Toast while the goals have already loaded.  

A better way is to have a default DOM element that shows 'Automatic scheduling of your goals in progress...' - and then replace that with the schedule once it's available.  

So instead of showing the empty list of days, only show the message in the middle of the screen.

tijlleenders (changes requested):
Yup, this is nicer.

Can you remove the italics (makes it harder to read)?
Also, please shorten the sentence to "Auto-scheduling...".

Then we're good to go!

tijlleenders (approved):
Perfect!

src/pages/MyTimePage/MyTimePage.scss
```diff
@@ -49,3 +49,12 @@
   left: 4%;
   border-radius: 8px;
 }
+
+.scheduling-message {
```
Tushar-4781, line 53: please use a custom class from short.scss for the font weight and font size prop

Tushar-4781 (changes requested):

tijlleenders:
@Tushar-4781 *bump*

Tushar-4781 (approved):

PR #2033, opened 2024-08-27
fix bugs related to deleted and archived goal items

### LINKED ISSUE #1991: Restoring from done missing
It's very annoying when you want to restore from done, and it's missing.

Implement restore from done, just like it is done in the main branch, adapt if needed:  
image

None.

**Additional context**
None.

vinaybadgujar102:
resolves #1991

### DIFF
--- a/src/components/GoalsComponents/MyGoal/MyGoal.tsx
+++ b/src/components/GoalsComponents/MyGoal/MyGoal.tsx
@@ -5,6 +5,8 @@ import { darkModeState } from "@src/store";
 import { ILocationState, ImpossibleGoal } from "@src/Interfaces";
 import { extractLinks } from "@src/utils/patterns";
 import { useParentGoalContext } from "@src/contexts/parentGoal-context";
+import { extractLinks, isGoalCode } from "@src/utils/patterns";
+import useGoalActions from "@src/hooks/useGoalActions";
 import GoalAvatar from "../GoalAvatar";
 import GoalTitle from "./components/GoalTitle";
 import GoalDropdown from "./components/GoalDropdown";
@@ -21,6 +23,7 @@ const MyGoal: React.FC<MyGoalProps> = ({ goal, dragAttributes, dragListeners }) 
 
   const [expandGoalId, setExpandGoalId] = useState("root");
   const [isAnimating, setIsAnimating] = useState(true);
+  const { copyCode } = useGoalActions();
 
   useEffect(() => {
     const timer = setTimeout(() => {
@@ -47,11 +50,17 @@ const MyGoal: React.FC<MyGoalProps> = ({ goal, dragAttributes, dragListeners }) 
     }
   };
 
-  const handleGoalClick = () => {
+  const handleGoalClick = (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+    e.stopPropagation();
+
     const url = extractLinks(goal.title);
     if (url) {
       const finalUrl = url.startsWith("http://") || url.startsWith("https://") ? url : `https://${url}`;
       window.open(finalUrl, "_blank");
+    }
+    if (isGoalCode(goal.title)) {
+      copyCode(goal.title);
+      return;
     }
     const newState: ILocationState = {
       ...location.state,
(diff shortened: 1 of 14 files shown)


tijlleenders (approved):
Yes this is nice.  
If you can, add the new goal 'pling' sound when restoring the goal.

vinaybadgujar102:
> Yes this is nice. If you can, add the new goal 'pling' sound when restoring the goal.

done

tijlleenders (changes requested):
Great.

I just found a bug:
[Screencast from 2024-08-28 16-57-58.webm](https://github.com/user-attachments/assets/622712fc-e764-4d1b-9c1b-5740479132b8)

The first time around is fine, but the second time around I press the completed Goal circles (no visible immediate effect). That apparently messes things up.

I now realize pressing the goal circle when it's completed (but not yet moved to Done) should have the same effect as if it's already in Done. This could be another issue, up to you.

src/components/GoalsComponents/MyGoal/MyGoal.tsx
Tushar-4781, line 1: can you please test collaboration? because if you receive two changes and then at the time of accepting the first change we move on to the second after second the notification dot should disappear. this behavior might get infected with the memo hook.

vinaybadgujar102: I tested the collaboration and found that the dot doesnt disappear after accepting the change even if MyGoal is not memoized. also there is a bug where collaborated goal appear both as normal and deleted goal. So if I try to open goal options of that collaborated goal I get modal of both deleted and normal goal action options. Thus it is bit hard to compare changes with the develop branch.

Tushar-4781: Noted

Tushar-4781:
@tijlleenders Please check the delete functionality again :) 
@vinaybadgujar102 can you please try collab again now. I've updated the PR

Tushar-4781:
Previously deleting a goal was taking too long because of the sound.

tijlleenders (changes requested):
When I complete multiple goals in Daily routines then select the circle on one of them, nothing happens.

Tushar-4781:
@tijlleenders what are you expecting in that case ?

tijlleenders:
@Tushar-4781 The same behavior as when it would be in 'done'. This is currently working for root goals - but not for subGoals.

tijlleenders:
The use case is if you mark something as done - then change your mind. You don't want to have to navigate away first.

vinaybadgujar102:
@Tushar-4781 the dot is correctly disappearing now after accepting changes according to your comment. Additionally I encountered some bugs related to share functionality for which I will create issues once I confirm them.

> When I complete multiple goals in Daily routines then select the circle on one of them, nothing happens.

@tijlleenders, I fixed this in latest commit.

Tushar-4781:
@vinaybadgujar102 I think there's a more efficient way to handle this. Since the component already has access to the length of the goals, it might be better to use that directly rather than passing it in as props

tijlleenders (changes requested):
Wow we should probably get some tests for all this behavior. Every round of manual testing we find something new. This is going to be hard to catch regressions on without tests.

I noticed now that crossing out a done item and then deleting an item will move the done item to done while moving the deleted to trash.  

I would expect only the deleted item moving to trash, and the done being crossed out untill you navigate away and back again.

vinaybadgujar102:
@tijlleenders, I agree. Creating tests for core features will help minimize redundant efforts in fixing regressions introduced by new changes.

Tushar-4781:
@tijlleenders can we have test in seperate pr

tijlleenders:
> @tijlleenders can we have test in seperate pr

Yes, of course.

tijlleenders:
> I noticed now that crossing out a done item and then deleting an item will move the done item to done while moving the deleted to trash.
> 
> I would expect only the deleted item moving to trash, and the done being crossed out untill you navigate away and back again.

[Screencast from 2024-09-02 12-29-00.webm](https://github.com/user-attachments/assets/2aab448d-688c-4820-aea4-f8290ffd99c4)

Tushar-4781:
@vinaybadgujar102 are u fixing it or do you want me to fix it ?

vinaybadgujar102:
> @vinaybadgujar102 are u fixing it or do you want me to fix it ?


You can fix it

Tushar-4781 (approved):

tijlleenders:
Holding off on 'fixing' this as state update exceptions on default React get complicated.

PR #2049, opened 2024-09-22
Tijl/ /fix playwright

tijlleenders:
The Playwright tests haven't been running for a while. That's just not right.  
Hope this fixes it.


vinaybadgujar102:
@tijlleenders, The tests are back to green! ✅

tijlleenders:
🎉🎊!!!

Can you explain the issue?

vinaybadgujar102 (dismissed):

vinaybadgujar102:
Firstly, test was very flaky and would fail at various different points, mostly while waiting for request/response and interacting with modals. I tried different techniques like using retries, waiting for network to be idle, and waiting for modal to appear. I also made the tests run in headed mode in CI/CD. Additionally, I tried different assertions to check if the shared goal was being updated correctly and chose the most robust one.

tijlleenders:
@vinaybadgujar102 That's great. Well done!

What is the best methid you chose?  
Can we go (or are we already) back to headless mode?

vinaybadgujar102:
> What is the best methid you chose?
Its situation based like for the modal thing, wait for the modal to be visible on viewport then do next steps.
Earlier we used to share the first goal from the starter goals. Now new goal is created and then shared between the users to remove the assertion conflicts which were arising previously related to goal titles.

>  Can we go (or are we already) back to headless mode?
We are currently running in headed mode only. I will check if we can move back to headless mode.

vinaybadgujar102:
@tijlleenders headless is also working fine

tijlleenders:
Ok, great!
We can merge then.

vinaybadgujar102 (approved):

PR #2060, opened 2024-10-13
feat: open goal modal with spacebar key

### LINKED ISSUE #2053: Open modal for goal with 'spacebar' key.
It's very annoying when you're used to using 'spacebar' to click on things - and it doesn't work.

When navigating through goals using the keyboard, pressing 'spacebar' key will go into the highlighted goal, just like tapping/clicking.

Using 'enter' key - but this is reserved for quick-adding goals.

**Additional context**
None.

harshalranjhani:
the goal modal now opens with spacebar key, a similar action which could be performed by clicking the goal as well

fixes #2053 

Screenshot:
image

### DIFF
--- a/src/components/GoalsComponents/GoalSublist/components/GoalHistory.tsx
+++ b/src/components/GoalsComponents/GoalSublist/components/GoalHistory.tsx
@@ -24,7 +24,6 @@ const breadcrumbStyle: React.CSSProperties = {
 };
 
 const BreadcrumbItem = ({ goal }: { goal: ISubGoalHistory }) => {
-  console.log(goal.goalTitle);
   const { t } = useTranslation();
   return (
     <span
--- a/src/components/GoalsComponents/GoalSublist/components/GoalHistory.tsx
+++ b/src/components/GoalsComponents/GoalSublist/components/GoalHistory.tsx
@@ -24,6 +24,7 @@ const breadcrumbStyle: React.CSSProperties = {
 };
 
 const BreadcrumbItem = ({ goal }: { goal: ISubGoalHistory }) => {
+  console.log(goal.goalTitle);
   const { t } = useTranslation();
   return (
     <span
--- a/src/hooks/useGoalSelection.ts
+++ b/src/hooks/useGoalSelection.ts
@@ -14,6 +14,7 @@ export const useGoalSelection = (goals: GoalItem[]): GoalItem | undefined => {
   const downPress = useKeyPress("ArrowDown");
   const rightPress = useKeyPress("ArrowRight");
   const leftPress = useKeyPress("ArrowLeft");
+  const spacePress = useKeyPress(" ");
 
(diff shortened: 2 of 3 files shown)


tijlleenders:
@harshalranjhani The original intent of the issue was to navigate 'into' the goal - instead of opening the modal...   
... but actually the way you naturally interpreted it probable makes more sense. Keep the left/right/up/down keys for navigation, and spacebar for 'action'.  


I'm good with it - except that the background navigates to a higher lever when one presses spacebar at a lower level:  
[Screencast from 2024-10-13 14-09-20.webm](https://github.com/user-attachments/assets/0a95e5f2-546a-478a-a3d5-3328760843ec)

harshalranjhani:
> background navigates to a higher lever when one presses spacebar at a lower level

hii, i've fixed this with the new commit. let me know if i fixed the issue correctly.

tijlleenders (approved):
Awesome, thanks!

harshalranjhani:
hey @tijlleenders , i was playing around with it when i realised that if the goal was a child goal, the state wasn't preserved and in the background, the parent goal title was disappearing, so i fixed that as well with this new commit and now everything should work perfectly fine!

tijlleenders (dismissed):
Good catch!

src/components/GoalsComponents/GoalSublist/components/GoalHistory.tsx
```diff
@@ -24,6 +24,7 @@ const breadcrumbStyle: React.CSSProperties = {
 };
 
 const BreadcrumbItem = ({ goal }: { goal: ISubGoalHistory }) => {
+  console.log(goal.goalTitle);
```
vinaybadgujar102, line 27: remove this log

src/hooks/useGoalSelection.ts
```diff
@@ -14,6 +14,7 @@ export const useGoalSelection = (goals: GoalItem[]): GoalItem | undefined => {
   const downPress = useKeyPress("ArrowDown");
   const rightPress = useKeyPress("ArrowRight");
   const leftPress = useKeyPress("ArrowLeft");
+  const spacePress = useKeyPress(" ");
```
vinaybadgujar102, line 17: can we use "Space" key code to detect the space key press?

vinaybadgujar102 (changes requested):

src/hooks/useGoalSelection.ts
harshalranjhani: @vinaybadgujar102 fixed it.

src/components/GoalsComponents/GoalSublist/components/GoalHistory.tsx
harshalranjhani: removed this as well.

vinaybadgujar102 (approved):

vinaybadgujar102:
Thanks for your contribution and effort 🙌

PR #2067, opened 2024-10-25
move goal feat

### LINKED ISSUE #1918: Move/Paste Goal/Budget in hierarchy

vinaybadgujar102:
resolves #1918


tijlleenders:
Looking good.  
Especially interested in the sharing behavior.  


Awesome to see you're tackling this complex issue!

src/components/GoalsComponents/DisplayChangesModal/ShowChanges.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <div className="move-info-item">
+          <span className="move-info-label">To</span>
+          <div className="move-info-value">{newParentTitle}</div>
+        </div>
+      </div>
+
+      {newParentTitle === "Non-shared goal" && (
+        <div className="warning-message">
+          <InfoCircleOutlined />
+          <span>The new parent goal is not shared. Changes will be ignored if accepted.</span>
```
tijlleenders, line 37: In my opinion, sharing should be per goal - independant of what the parent goal is.  

If I/you change the parent goal of our shared goal - that should not affect propagation of changes on our shared goal, right?

vinaybadgujar102: this is for the situation when child is moved out of the shared goal. the above code is stale. 

> If I/you change the parent goal of our shared goal - that should not affect propagation of changes on our shared goal, right?

this is handled differently. if we move the root shared alltogether then the suggestion will discarded altogether by the reciever.

tijlleenders: Yeah, what I'm saying is that maybe we should be dropping the whole concept of letting shared status depend on some 'root-goal'?  

In my view, when you share a goal - all descendants are also shared (or not with opt-out).  
From then on out - they are 'autonomous' from each other.  

Of course if you delete a tree - all descendants will also be suggested to be deleted - but not the ones you/partner have moved out of the tree previously.  


Basically, this makes sharing/deleting a tree the equivalent of individually sharing/deleting all goals in the tree. The advantage is that after that you are free to move things around individually without any extra complexity/dependency on other goals.

vinaybadgujar102: I agree that dropping root-goal dependency and treating each goal independently would make the system more flexible and intuitive.

I'll start implementing these changes.

vinaybadgujar102:
@tijlleenders when there is a change in shared descendant, where should the notification be displayed. root or the parent (if new goal added) / goal itself.

tijlleenders:
@vinaybadgujar102 The idea is that the notification dot travels up the tree to the highest level - just like now.  

A possible difference with the current situation is that the dot should not care about sharing when 'travelling up' the tree. It is also realtime (not a property of ancestors) - as things are always moveable.  

Does this make sense?   
Do you see any problems?

vinaybadgujar102:
@tijlleenders please do an initial review. I'm planning to integrate LWW later, once the move functionality is more robust. let me know if UI/UX of the move buttons requires improvement (you can ignore the icons currently used on the move buttons).

tijlleenders:
A confirmation after the move has completed is also nice to have.

tijlleenders:
- When I move a shared goal into a non-shared goal, the partner doesn't see the shared goal any more. If I move it back into a previously shared goal, it appears again.

You're getting there, though!

vinaybadgujar102:
@tijlleenders made the changes

tijlleenders:
Awesome work @vinaybadgujar102 !

I confirm the goal doesn't disappear anymore on receiver side.  

Yeah this user experience is much better - and I like the permanent message at the top - instead of a temporarily visible Toast/Notification.  

Some suggestions:  
- Use the popup with the "Don't show again" - just like when completing a Goal for the first time - instead of introducing a new UI element (with new styling).
- Split the long sentence into two short sentences: 
  - Navigate to the new location
  - Press the + button to move.
- Pressing the + once when in the 'move' state should show the menu. Nobody will know they need to long press - and it's also harder to long-press. There should be a 'Cancel move' in the + button menu too to get out of this state and return to the quick-add goal on single tap of +.

vinaybadgujar102:
> Use the popup with the "Don't show again" - just like when completing a Goal for the first time - instead of introducing a new UI element (with new styling).

So when user clicks the move action then a confirmation popup should appear with text

- Navigate to the new location.
- Press the + button to move. 

Then single tap on + will show two menu
1. Move here
2. Cancel move

@tijlleenders Right?

tijlleenders:
@vinaybadgujar102 
Yes, that sounds better, actually.

vinaybadgujar102:
@tijlleenders made the requested changes

tijlleenders:
Great, some minor corrections:  
- Split the long sentence into two short sentences:
  - Navigate to the new location.
  - Press the + button to move.

- Don't Capitalize Every Word
  - Move here
  - Cancel move

- Move mode should stay active until you press cancel or actually move. 
  Now if I press +, then click somewhere else than the two options, and return, the move options are gone.

tijlleenders:
@vinaybadgujar102 is this excluding LWW?

vinaybadgujar102:
> @vinaybadgujar102 is this excluding LWW?

I have implemented the function to handle the conflict resolution just have to integrate it with the existing code.
Should i do it in this PR only or separate?

tijlleenders:
Yes, please implement the sharing with LWW too, if it's ready to add.  

Currently, in this branch - if I rename a shared goal - it completely disappears from the other persons' partner page. The moving works fine, congrats!

vinaybadgujar102:
@tijlleenders, integrated the LWW for collaborated goals for all actions excluding change of adding new subgoals to shared goals and movement of non shared goals to collaborated goals.  Check if it is working correctly?

Is there need to implement lww on partner goals?

tijlleenders:
@vinaybadgujar102 This is AWESOME!!!

tijlleenders (approved):

tijlleenders (changes requested):
I have some issues with the default goals being populated sometimes. Do you experience any issues after clearing memory?

[Screencast from 2024-11-15 08-55-28.webm](https://github.com/user-attachments/assets/6f636d3d-6dff-40b8-a562-ad8fd6f9270c)

vinaybadgujar102:
> I have some issues with the default goals being populated sometimes. Do you experience any issues after clearing memory?

Yeah. Sometimes only 2-3 starter goals are created after memory refresh. Its not introduced in this pr, I have been experiencing it from long time ago

tijlleenders (dismissed):
Ok, please add an issue for that then.  
Super excited for this new move feature, thanks!

tijlleenders:
@Tushar-4781 Can you review the code?

Tushar-4781 (changes requested):
please cut down css. use the short classes defined in shot.scss

Tushar-4781:
1. After clicking the Collaborate button for a goal in partner mode, it neither navigates to the My Goals page nor removes the goal from the current view.
2. Adding a sub-goal in collaboration mode results in duplicate entries appearing in the Display Changes menu.
3. When a sub-goal is added under another sub-goal and subsequently moved one level higher, it moves successfully, but the circle icon still indicates the presence of sub-goals.
4. All sub-goals under the root goal (the one being shared in collaboration) incorrectly display the participants' icon. Only the root goal itself should display the participants' icon.

Tushar-4781 (changes requested):
1. After clicking the Collaborate button for a goal in partner mode, it neither navigates to the My Goals page nor removes the goal from the current view.
2. Adding a sub-goal in collaboration mode results in duplicate entries appearing in the Display Changes menu.
3. When a sub-goal is added under another sub-goal and subsequently moved one level higher, it moves successfully, but the circle icon still indicates the presence of sub-goals.
4. All sub-goals under the root goal (the one being shared in collaboration) incorrectly display the participants' icon. Only the root goal itself should display the participants' icon.

tijlleenders:
Good catch!  
I didn't test the collaboration part - only the sharing.  

For 4. I disagree. Since goals are shared individually now, instead of all being reliant on a 'root' goal, all goals that have sharing enabled should show this. I realize this can get a little overwhelming in the UI - but that's something we'll have to solve (ie by changing the circle icon to a square for non-shared goals - and then getting the 'followers' details via the edit menu instead of directly on the goal).

tijlleenders:
@Tushar-4781 @vinaybadgujar102 I suggest we keep it out of this (already massive PR) and log a separate issue for it.

vinaybadgujar102:
> Adding a sub-goal in collaboration mode results in duplicate entries appearing in the Display Changes menu.

@Tushar-4781, I think this is happening due to strict mode

vinaybadgujar102:
@Tushar-4781 ready for review

Tushar-4781 (changes requested):
1. Share a goal with a partner
2. create a subgoal "Subgoal 1" and send then refresh on both sides.
3. now create another subgoal inside subgoal 1 right after creation move it one level up with "subgoal 1" then refresh on the other side. the goal is not moved it is still in subgoal1

vinaybadgujar102:
@Tushar-4781  do u have any suggestions to solve it. its working fine in development but not in production. is there a way to test production build on local setup

Tushar-4781:
@vinaybadgujar102 can you resolve these conflicts before i give a shot

vinaybadgujar102:
@Tushar-4781 fixed the conflicts

Tushar-4781:
@vinaybadgujar102 in local the my goals page is constantly being re-rendered. are you aware of this issue or the LOC(s) that are causing this ?

vinaybadgujar102:
@Tushar-4781 u can review now. for now I removed the merge conflict resolution logic cause its introducing lot of complexity to already complex pr. i will add it back once core move feature is stable enough.

Tushar-4781:
@vinaybadgujar102  2 Failing cases that i came across 

1. You share a goal with person A
2.  person A make a suggestion
3. accept the incoming suggestion
4. refresh on person A side the suggestion is not added but your side the goal is added 

1. You start collaboration on a goal
2. You create a subgoal 1 and  create a subgoal 2 inside the  subgoal 1 
3. now refresh now on both sides
4. now  you move the subgoal 1 into root after this, now you make modification on subgoal 2
5 now refresh at person A side and reject the move 
6.  the updates after moving are not shown in person A side

vinaybadgujar102:
Hey @tijlleenders,  I tested the move functionality with most scenarios I could think of but I want to make sure I am not missing anything. Do you have any edge cases in mind that this should handle?

tijlleenders:
@vinaybadgujar102 No I don't have any other specific scenario's in mind right now. I think me using it with one or two people in practice on the dev release site is the best approach.

tijlleenders:
When A move a new (previously unshared) goal into a collaborating goal with B, the message says it's a move. Technically it doesn't matter. A move or not - all B knows is that there is a new Goal created by A. How do we know it was moved - and why do we make the effort of making a distinction? It seems this overcomplicates the logic for no real benefit. I would be fine not making the distinction, unless I'm missing something.

image

tijlleenders:
Glitch on creating a new goal inside a previously shared goal and then trying to move it.

[Uploading Screencast From 2025-01-18 16-34-07.webm…]()

tijlleenders:
Not getting a notification for move after creating a subgoal inside a previously shared (not collaborated) goal: 
[Screencast From 2025-01-18 16-34-07.webm](https://github.com/user-attachments/assets/93481f13-79f1-4716-abd2-9d9d795a80d6)

tijlleenders:
> @vinaybadgujar102 2 Failing cases that i came across Case 1
> 
>     1. You share a goal with person A
> 
>     2. person A make a suggestion
> 
>     3. accept the incoming suggestion
> 
>     4. refresh on person A side the suggestion is not added but your side the goal is added
> 
> 
> Case 2
> 
>     1. You start collaboration on a goal
> 
>     2. You create a subgoal 1 and  create a subgoal 2 inside the  subgoal 1
> 
>     3. now refresh now on both sides
> 
>     4. now  you move the subgoal 1 into root after this, now you make modification on subgoal 2
>        5 now refresh at person A side and reject the move
> 
>     5. the updates after moving are not shown in person A side

Great catch.

Case 1:  
**The moving still doesn't work after not accepting a suggestion.** The goalId is still the same so move should continue to work - even if the titles don't match?

Case 2:  
It's better to use "person A" and "person B" instead of 'you'. I had trouble understanding this case. This one seems to work now, although the order of presenting the changes flipped.

tijlleenders:
@vinaybadgujar102 My general impression is that this is still very brittle. Let's investigate the issues above first and then see where we stand.

vinaybadgujar102:
@tijlleenders i agree. Users dont care about how a new subgoal is added, so its just a simple replacement for header title in changes modal.
The 'newgoalmoved' change is slightly different from 'subgoals' change in terms of business logic. For the former, we also need to send updates for its descendants in a single request only. Additionally its better to have seperation of concerns for any further future updates.

tijlleenders:
@vinaybadgujar102 Maybe we should talk this through to make sure we understand. I don't mind batching changes logistically - but spearate changes should be separate events, logically.  

A move is not the same event as a title change and they can't be merged into one event - but should stay independent so even out-of-order arrival on the other side would result in the same end-state.

tijlleenders:
@vinaybadgujar102 The comment about 'sending descendants' worries me. Only goals that have actually changed should be sent. The 'root goal' is purely for notification purposes on the receiver side - and is not part of 'the conversation'. It's up to the receiver to figure out if that field needs changing.  

I would suggest to rename this (private) property to 'notification goal'. We could leave it out completely and dynamically 'calculate' it - but we chose an approach where we only (re)calculate it whenever something moves.

vinaybadgujar102:
@tijlleenders 
> The comment about 'sending descendants' worries me

I m discussing the scenario of moving goal that hasnt yet been shared with the receiver. When attempting to move such a goal and share those changes with participant, we first check if the goal is already shared. If it isnt we treat it similarly to the case where a new subgoal is added to shared parent goal. The key difference here is that a "non-shared moved goal" might have many children and further descendants. In such cases, updates must also be sent for all these descendants as a new subgoal.

While its possible to send individual requests for each descendant, this could lead to too many requests if there are many subgoals involved.

For the second scenario, where the goal being moved is already shared, we can treat it as a "modified change," as you suggested. In this case, we currently only send a change update for that specific goal being moved, without including any descendants.

I will also address the "root goal" aspect as you mentioned

tijlleenders:
@vinaybadgujar102 Thanks for clarifying. I think we are good! The descendants can be sent in one (or better batched) in https request(s) - but please do keep them as separate logical events.  

Why? Down the line we will move away from assuming that ALL descendants are automatically shared together with the explicitly shared goal - or at least give an opt-out for some descendants. 

There are some UX implications to figure out. This will be separate from this PR. I think showing private goals as squares and shared goals as circles could be another step towards 'atomic' sharing (vs the abandoned approach of sharing whole trees as one unit).

tijlleenders:
@vinaybadgujar102 Can you summarize the status of remaining tasks for this PR?

vinaybadgujar102:
@tijlleenders just integration of LWW resolution and renaming of rootGoalId is remaining. will complete today only

vinaybadgujar102:
@tijlleenders please test the feature again.  now treating move as same change event as modifying goal and move suggestion working for partner goals

tijlleenders:
Great, thanks!  

Found an issue:  
A shared goal should still remain shared, even if its new parent after moving isn't:  
[Screencast From 2025-02-14 09-05-13.webm](https://github.com/user-attachments/assets/44fe2920-edef-446b-be31-fd1a2fec2828)

tijlleenders (changes requested):
Shared goal disappearing after moving into non-shared parent.

vinaybadgujar102:
> Shared goal disappearing after moving into non-shared parent.

@tijlleenders, just to be clear, moving into a non-shared goal should not be acknowledged by receiver in both shared and collaborative goals, right?

tijlleenders:
> @tijlleenders, just to be clear, moving into a non-shared goal should not be acknowledged by receiver in both shared and collaborative goals, right?

Interesting... I haven't thought this through...  

Right, in principle there is nothing to acknowledge - since you can't communicate the new (non-shared) parent to the other person.  

The situation would be different if the new parent was also shared.  

But the only question can still be asked if we want to commjnicate the next best parent/ancestor that is shared... I think giving as much info as possible should be the way to go, so whenever a shared goal is moved to a non-shared parent we should communicate the first shared ancestor as the new parent - or root if none are found.  

Does this help?

vinaybadgujar102:
> whenever a shared goal is moved to a non-shared parent we should communicate the first shared ancestor as the new parent - or root if none are found.

@tijlleenders, Im struggling to come up with a testing scenario for this. Could you suggest one?

tijlleenders:
Sure, I understand having a clear use-case helps.  
  
Example 1:  
- Work => Project A with technology XYZ => Stuff to do  (all shared with my work boss / team)
Now let's say I have a secret wish to work at another company that requires me to have experience on technology XYZ. I could then have a personal goal below shared 'Work' goal to 'Gain XYZ experience for switching jobs'. I would not like to share that with my boss / colleagues. The situation then becomes:
- 'Work' (shared) => 'Get experience XYZ to switch jobs' (private) => 'project A' (shared) => 'stuff to do' (shared).  

If a new colleague comes along and I share my 'Work' and 'project A' goals with them - they should be able to see that 'project A' is a subgoal of my 'Work' goal - even though the immediate parent of 'project A' is non-shared. Otherwise they would end up next to each other in the view where the colleague looks at my goals.

vinaybadgujar102:
@tijlleenders, but we are currently storing participants in whole goal tree. if we want to support mentioned functionality then we have prompt user if they want to share newly added descendant with a specific partner or not.

tijlleenders:
> @tijlleenders, but we are currently storing participants in whole goal tree. if we want to support mentioned functionality then we have prompt user if they want to share newly added descendant with a specific partner or not.

For 
- Goals to be **_shared_** atomically  
- shared and private goals to be mixed together in a hierarchy,  
  the participant information needs to be available on a per-goal level, not a tree of goals.  

The only reason for using a tree is so that the notification dot can be shown at the right level higher in the tree - instead of having to dynamically traverse up till the highest level that is still shared.  

You could still choose to use the 'root goal' tree indicator for ease/speed/efficiency - but it would be on-top of the information available on the goal.  
  
In any case I would suggest to rename the 'root goal' to 'notification goal' - as that is more aligned with it's purpose/meaning.

vinaybadgujar102:
@tijlleenders, renamed "rootGoalId" to "notificationGoalId" accordingly.

My query remains unresolved. I can't think of a way to achieve this state:

Work (shared) → Get experience XYZ to switch jobs (private) → Project A (shared) → Stuff to do (shared).

Since "Work" is shared, any new subgoal added to it triggers an event sent to all participants. What I m suggesting is that we first need to implement functionality allowing the sender to select which participants should receive the updates

tijlleenders (changes requested):
Would be nice to land this BIG feature!  

Please document the tests you have included to test this with collaboration.

vinaybadgujar102:
### **Partner Mode and Collaboration Mode:**

1. **Moving a new goal under a shared parent** and then sharing that moved goal works as expected.
2. **Suggesting move changes in Partner Mode** functions properly, even when a private goal is in the ancestor hierarchy.
3. **Moving a shared subgoal to the root** works without issues.
4. **Moving goals across shared goals** functions as expected.
5. **Propagating accepted changes to all participants** works correctly.

This feature is mostly stable based on my initial testing. However, a second review is needed to address any edge cases that may arise.

### **Collaboration Mode Testing:**

In order to fully ensure robustness, this feature should be used regularly in a daily routine context within Collaboration Mode for thorough testing.

---

### **Query Regarding LLW Conflict Resolution:**

Currently, we are not using our basic LLW approach. If necessary, I can add it back, but the current method does not resolve conflicts on a per-field basis. Instead, we accept the latest change, even if unrelated fields are updated. To improve this, we should consider resolving conflicts on a per-field basis so that other field changes do not get removed when unrelated fields are updated.

vinaybadgujar102:
@tijlleenders is screen recording required?

tijlleenders:
Excellent, yes a recording would be nice and is straightforward if the tests are automated.  

Af for the field level conflict resolution: thatvisba separate feature which requires timestamps per field to avoid storing all versions in combination with rather complicated comparison logic.  

The added value is low at this moment so I don't think we'll be implementimg this any time soon.

vinaybadgujar102:
@tijlleenders, attached the automated test recordings

Testing move with shared goal:
https://github.com/user-attachments/assets/dc722d8d-c157-4e11-81aa-a79d14dc7dc6

Testing move with shared goal and private goal in between:
https://github.com/user-attachments/assets/da0759ab-4fee-455d-9c28-60db9f25f3ff

Testing move with collaborated goals between 3 users:
https://github.com/user-attachments/assets/722a92a7-e871-435d-ae73-ecdbafb98600

tijlleenders (changes requested):
@vinaybadgujar102 Nice work. We can merge after conflicts are resolved.

tijlleenders (approved):

PR #2086, opened 2024-11-30
allow deleting contacts

### LINKED ISSUE #1790: Allow deleting contacts
It's very annoying when you have test contacts polluting your contact list. They appear on My partner page and also in the sharing menu. When I change devices I need to add new contacts for resetting the sharing - but the old contacts remain.

Change the way users are shown from a horizontal list to a vertical list, similar to a Goals list. Have the contact initials appear in the circle. Tapping the circle brings up an Edit name or Delete modal.

- Allow removing a contact by long pressing and showing an option modal with edit and remove.  
- Use the switch partner mode button => but then how to switch back?
- Use the header. How would you know?
- Use the three-dot menu. That would surprise me.

These are all less intuitive.

**Additional context**
How it looks now:

How it should look:

vinaybadgujar102:
Resolves #1790


vinaybadgujar102:
@tijlleenders what should happen on receiver side when a contact is deleted by a user?

tijlleenders:
> @tijlleenders what should happen on receiver side when a contact is deleted by a user?

Interesting question... I see three possible reasons for deleting a contact(device):
- that contact(device) no longer exists
- the relationship is broken bilaterally (both parties don't want to share)
- the relationship is broken unilaterally (only one party doesn't want to share)
Only in the last case there is a reason to:
- maintain the relationship id
- possibly inform the other party they are being 'ghosted'

For now, let's not add any extra requirements.  
Deleting is one-sided and doesn't affect the other party or the relationship id.

tijlleenders:
I think we can remove the 'blocked' label?

tijlleenders (changes requested):
This is pretty cool, nice work!  

There is a small glitch in the way shared subgoals are being handled:  
[
[Screencast From 2025-02-22 10-37-29.webm](https://github.com/user-attachments/assets/a05f6560-d8c0-410f-b963-5bb5f2cd02f9)
](url)

vinaybadgujar102:
> This is pretty cool, nice work!
> 
> There is a small glitch in the way shared subgoals are being handled: [ [Screencast From 2025-02-22 10-37-29.webm](https://github.com/user-attachments/assets/a05f6560-d8c0-410f-b963-5bb5f2cd02f9) ](url)

@tijlleenders, I cant reproduce it. Could you please provide details on how you arrived at this state of goals in partner mode?

tijlleenders:
@vinaybadgujar102 Maybe it's the commit you did, but I can't get a sharing link anymore.

tijlleenders (changes requested):
Sharing link not working so can't test or try to reproduce the issue.

tijlleenders:
@vinaybadgujar102 Can you check what's failing the Playwright test?  
I'd like to keep dev green.

vinaybadgujar102:
@tijlleenders, fixed the test and please check again if this feature is correctly

tijlleenders (approved):
Nice!

Also like the toast for when there are no contacts remaining + redirect.

tijlleenders:
@Tushar-4781 Can you do tech review on this?

PR #2106, opened 2025-01-23
remove taskdone collection and send goal completion stats to scheduler

### LINKED ISSUE #2103: Remove tasksDoneToday collection
In #1407 we implemented tasksHistory and tasksDoneToday collections, but the information from tasksDoneToday can also be queried from taskHistory - so it is actually not needed.

Remove the tasksDoneToday collection - and send the required data for the scheduler from taskHistory:

For each goal that is sent to scheduler, filter/aggregate it down to:
- total_time_completed (for reducing the tasks required for simple goals)
- tasks_completed_since_Monday (necessary for reducing budgets, but don't filter on budget goals as the simple goal could be a subGoal of a budget Goal - use the scheduled datetime, not the completed datetime) - this also includes completed Tasks with scheduled datetime today or in the future - so appropriate day-budgets can be reduced today or in the future.

Add these fields to the goal being sent to the scheduler.

Reducing the number of hours of the goal.  
A. This is not correct. The goal represents the ambition, not the realization.
B. An option could be to include it as attribute(s) of the Goal (i.e. realizedHours). Keeping that accurate could get messy, due to the sliding window - so querying the tasksHistory table directly at schedule query time dynamically is simpler, without the indirection of the Goals collection.

**Additional context**
None.

vinaybadgujar102:
Resolves #2103


tijlleenders:
Why do we store the TaskId?

vinaybadgujar102:
> Why do we store the TaskId?

mostly used to strikeout the specific completed task in mytime page

vinaybadgujar102:
@tijlleenders please check if this working as expected

tijlleenders (changes requested):
Can you clean up the console logs that are not necessary?

tijlleenders (changes requested):
Please remove the 'tasksCompletedToday' at the root of the JSON being sent to the scheduler - and any corresponding code.  

The information is now present on every Goal - so this is no longer required.

tijlleenders (changes requested):
`totalDurationCompletedOrSkipped` can be simplified to just include the amount of completed hours, so `totalDurationCompleted`. For a habit, it the total of skipped doesn't matter. We only need the amount skipped since monday.

`tasksCompletedOrSkippedSinceMonday` also needs to be split in `tasksCompletedSinceMonday` and `tasksSkippedSinceMonday` so that we can make that distinction in the ouput.  

The scheduler will then be able to generate the complete schedule, including taskstatus - so no more need to remember taskid/status on the frontend.  

So for a Goal that has no hours skipped, but a few completed, this is expected:  
```
"stats": {
        "totalDurationCompleted": 5,
        "tasksCompletedSinceMonday": [
          {
            "scheduledStartDateTime": "2025-03-21T00:00:00",
            "duration": 5
          }
        ]
      }
```

For a Goal that has some hours skipped in the past, but also some since Monday:  
```
"stats": {
        "totalDurationCompleted": 3,
        "tasksCompletedSinceMonday": [
          {
            "scheduledStartDateTime": "2024-04-26T14:00:00",
            "duration": 3
          }
        ],
        "tasksSkippedSinceMonday": [
          {
            "scheduledStartDateTime": "2024-04-26T07:00:00",
            "duration": 5
          }
        ]
      }
```

vinaybadgujar102:
> Please remove the 'tasksCompletedToday' at the root of the JSON being sent to the scheduler - and any corresponding code.
> 
> The information is now present on every Goal - so this is no longer required.

scheduler is throwing error after removing this field. is this expected?

tijlleenders:
@vinaybadgujar102 Yes, it's expected that the current scheduler will not like you changing the interface. I'll be adding a new version of the scheduler that can handle the changed input to this PR later.

tijlleenders:
Also... after I do that we'll need to make some additional changes to process the new output, that will include a status integer for todo (0), done (1) or skipped (2).

tijlleenders:
@vinaybadgujar102 Let me know when you're done so I can get the scheduler upgraded.

tijlleenders:
@vinaybadgujar102 I updated the scheduler to expect the new interface.
Whenever a Task is __done__, it will include a `status` field with value `0`.  
Whenever a Task is __todo__, the status field is omitted - as that's the expected situation.  

Whenever a Task is __skipped__, this will not be shown in the output - but the budget is reduced accordingly. This also means that the Task will 'disappear' from the calendar - to make way for other Tasks.

vinaybadgujar102:
@tijlleenders 

```javascript
"stats": {
        "totalDurationCompleted": 5,
        "tasksCompletedSinceMonday": [
          {
            "scheduledStartDateTime": "2025-03-21T00:00:00",
            "duration": 5
          }
        ]
      }
```

the scheduler is breaking when tasksSkippedSinceMonday or tasksCompletedSinceMonday is not passed to stats object

tijlleenders:
I'll release a version of the scheduler where they are optional.

tijlleenders:
@vinaybadgujar102  
Can you omit the whole stats object if there are no hours completed, and no skipped hours?

tijlleenders (approved):
Awesome!

PR #2128, opened 2025-04-01
fix schedule not cached properly

### LINKED ISSUE #2125: Schedule not cached any more
**Describe the bug**
When switching from MyGoals to MyTime, the scheduler is called with an 'Auto-scheduling...' message.

**To Reproduce**
1. Go to MyGoals
2. Go to MyTime
3. See 'Auto-scheduling' message

**Expected behavior**
- Schedule should be instantly available from cache.
- Cache recalculation is triggered each time a scheduler-relevant goal is modified.
  Careful: ONLY goals that have/had a duration before/after modification trigger a re-schedule / cache update

**Screenshots**
If required.

**Additional context**

vinaybadgujar102:
resolves #2125

### DIFF
--- a/src/Interfaces/IScheduler.ts
+++ b/src/Interfaces/IScheduler.ts
@@ -84,3 +84,5 @@ export interface ISchedulerInput {
 }
 
 export type TBufferValue = { nextBuffer: number; availableBuffer: number };
+
+export type SchedulerCacheCode = "not-exist" | "expired" | "found";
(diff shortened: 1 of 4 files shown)


tijlleenders (approved):

tijlleenders:
@vinaybadgujar102 Why do we still see the 'Autoscheduling...' wait screen if the schedule can be loaded near-instantly from cache?

vinaybadgujar102:
> @vinaybadgujar102 Why do we still see the 'Autoscheduling...' wait screen if the schedule can be loaded near-instantly from cache?

@tijlleenders, even though we are storing scheduler response as cache in db. but to check if this response is available in cache we do some async operation to generate the key for comparison. This leads to some delay.

tijlleenders:
I understand that's how we do it now.  
ainpropose removing the check, or doing it after loading the screen - since the only time it won't be there is on a fresh install. After every change to goals/tasks that impacts the schedule we recalculate and cache anew, so this should always be up-to-date.

PR #2142, opened 2025-05-19
feat: use react query for goal related crud

vinaybadgujar102:
Time taken to show newly added goal the moment addgoal function is called (CPU: 20x slower): 

Current:
Screenshot 2025-05-19 at 10 54 58 PM

Updated:
Screenshot 2025-05-19 at 10 54 41 PM


tijlleenders:
I noticed the goal adding was relatively slow - but thought it was a regression.  
All actions should feel snappy - as there is no technical reason for latency in a local-first app.

vinaybadgujar102:
@tijlleenders, most goal related actions are optimistically updated on ui first

tijlleenders (approved):
Very nice.  
I noticed the scheduler is not cached after adding a new goal with duration, causing it to recalculate when switching to MyTime.  
Not sure - but it looks like a regression. If not let's file it as a separate issue.

tijlleenders (approved):

Issue #1584, opened 2023-10-31
Enable after before before in budget time of day slider


tijlleenders:
**Is your feature request related to a problem? Please describe.**
It's very annoying when I can't schedule a budget to sleep between 10 in the evening and 6 in the morning.

**Describe the solution you'd like**
Let me :
- differentiate between start slider (with a >) and the end slider (with a <)
- AND allow me to drag the end slider before the start.
- When I do that : show a permanent warning "Ends on next day."

**Describe alternatives you've considered**
Auto extend the timeline to the next day. This is more confusing - and requires more complicated code than the proposed solution.

**Additional context**
None.

vinaybadgujar102:
> differentiate between start slider (with a >) and the end slider (with a <)
> AND allow me to drag the end slider before the start.

@tijlleenders, I checked the antd's docs but couldn't find way to do the above-mentioned modification on Slider Component. I think we have to build a custom slider component to meet our requirements.

@Tushar-4781 what are you thoughts on this?

tijlleenders:
@vinaybadgujar102 @Tushar-4781 Let's discuss this further when it's on the top of the Parking lot prio.

PR #2143, opened 2025-06-03
simplify goal config goal and show it in goal's page

### LINKED ISSUE #2108: Show budget edit modal 'on-page'
It's very annoying when you are a new user - or even as an experienced user - that you have to click the budget circle to edit a budget. It is not obvious to figure out, and it requires two clicks before you can edit in a modal with restricted space.

- When 'on the page of a goal', whenever you click on the description, uncollapse the contents of the edit modal 'on the page'. A modal is not necessary. The subGoals remain visible, just lower.
- If the Goal/Budget has no subGoals, show/uncollapse the edit modal 'on the page' directly, as there is plenty of space for it.

None.

**Additional context**
Please voice to double-check understanding before starting work on this issue.

vinaybadgujar102:
resolves #2108


vinaybadgujar102:
@tijlleenders, i have made the required changes as mentioned in the issue. You can drop a review. Once everything looks good on your end, i will check whats going wrong with the tests

tijlleenders (changes requested):
Fantastic work - this is so much better in terms of user experience!  

image

I think this should say 'Update budget' like it does when you use the circle to edit ...
... but now that it is already created and there is no popup - why not remove this button all together - and save it automatically?

So please remove the button and save the changes as they happen.

tijlleenders (changes requested):
1.  Please remove this as it is showing redundant info.  

image

2. When clicking to expand - make the whole area clickable so it is super easy to uncollapse.  Now I have to 'aim' at the 6 subgoals to get the Budget edit menu to uncollapse. Clicking anywhere else doesn't do anything.

image

tijlleenders:
@vinaybadgujar102 Just leave it empty for now. We can optimize when handling #2145

tijlleenders (changes requested):
Looking good!

When going into a goal/budget with no subGoals I expect the details to be uncollapsed - but without focus on the title.

tijlleenders (changes requested):
Please check navigation behavior:



Also, when uncollapsong goal/budget details - the title doesn't have to be auto-focused as this causes the keyboard to appear - which mostly is not what you want as the title is relatively stable.

tijlleenders (changes requested):
1.
The navigation behavior is different than on the main branch. When navigating the hierarchy, navigation doesn't get 'collapsed'. So if I go into GoalA>GoalB>GoalC, then navigate to GoalA>GoalB>GoalD and press back twice, I end up in GoalC instead of GoalA.

2.
When in an uncollapsed/expanded subGoal and navigating up, the parent goal then is also uncollapsed/expanded - when it shouldn´t be.

tijlleenders (changes requested):
Regression: When adding a goal/budget title field is not autofocussed. I expect to be able to immediately start typing when adding a new goal/budget.

Also, the area for collapsing goal/budget details seems to be limited to the top half of the area around the goal/budget circle color.

tijlleenders:
> @tijlleenders, so should this highlighted area also allow toggling?


Yes please.

tijlleenders (changes requested):
Nice!

Navigation works as expected.
Autofocus on new goal fixed.
Goal collapse area fixed.
Speed of navigation fixed.

Still to fix:
1. Regression: goals don't update automatically after an import.
2. Several regressions filed as separate issues that need to be fixed or marked as resolved by this PR so regressions don't go to production.

vinaybadgujar102:
@tijlleenders Could you please check if the color picker is now visible?
Also, I wanted to confirm that the newly added goal is not appearing at the top because the "impossible" goal is being prioritized.
To make that clearer, i have added "!" back as a prefix to the impossible goal’s title.

tijlleenders (changes requested):
Only checked what you asked:
- Color picker seems fine. What was the problem?
- Yes, adding back ! is a good one - I forgot about this.

Did see another todo:
- Hints toggle not being persisted when editing the goal 'on screen´ (as opposed to in the modal).

tijlleenders:
> @tijlleenders i checked it. this happens only for sleep budget. Sleep is the only one currently having beforeTime less than afterTime and our current slider doesnt support values like this. I think this can be fixed in #1584.  


Yes, I agree this is not in scope of this PR.  
Will merge this PR once Playwright tests are passing.

tijlleenders (commented):
Nicely done!

tijlleenders (approved):

Issue #2146, opened 2025-07-04
Enter on title doesn't save goal/budget


tijlleenders:
**Describe the bug**
When adding a new goal/budget - pressing enter on keyboard used to save the goal/budget.

**Additional context**
Regression only in mobile version. Web still works.

vinaybadgujar102:
fixed by #2150

PR #2150, opened 2025-07-10
fix: pressing enter not saving the goal

vinaybadgujar102:
#2146


tijlleenders (changes requested):
It's not working for me on mobile using DuckDuckGo browser or Vanadium browser.
OS: GrapheneOS (GOS) + builtin GOS keyboard

tijlleenders (approved):
Great!

PR #2158, opened 2025-09-04
Improve goal fetching performance

vinaybadgujar102:


tijlleenders:
Is there an index on the database?

vinaybadgujar102:
yeah, i was just testing some stuff. the newly added compound index isnt very useful, the advantage is minimal, and i will remove it

tijlleenders:
First thing is to measure where the lag is coming from - unless it is obvious & quick fix.

It feels like we rely on caching to make things look fast, whereas getting something from even a table with 100s of entries should not be a problem as it is local storage? Can caching be turned off to check/test non-cached performance?

tijlleenders:
Could it be the new component introduced? What happens if you temporarily skip that?

vinaybadgujar102:
@tijlleenders, please check if the goal fetching performance is improved

tijlleenders:
1.
On mobile it takes a long time for the goals to load after a refresh. 

2.
Seems like there is some kind of preloading/caching going on? The first time I enter a goal it lags (notice the visible tap before continuing. The second time it goes into the goal directly.

https://github.com/user-attachments/assets/d53a1fc0-94e5-4060-946b-f52647b491a9

vinaybadgujar102:
fixed in #2159

PR #2159, opened 2025-11-06
improve app performance

vinaybadgujar102:


vinaybadgujar102:
@tijlleenders could you check whether goal loading is performing better or still slow?

tijlleenders (dismissed):
@vinaybadgujar102 This is much better.
There is still an initial very noticeable black page on initial entry into a goal tree from the root level... but it is not as problematic as the current situation.

tijlleenders:
I see you decided to add tanstack router in favor of react router.  
Isn't tanstack used for server-query heavy apps? This app is supposed to be local-first ... with background updates. Isn't that a simpler way of achieving fast navigation of goals?

If the database is local - navigation should be instant.
Queries to the server are NOT required to be instant, so I'm wondering if this is the right approach.

tijlleenders:
> I see you decided to add tanstack router in favor of react router.  
> Isn't tanstack used for server-query heavy apps? This app is supposed to be local-first ... with background updates. Isn't that a simpler way of achieving fast navigation of goals?
> 
> If the database is local - navigation should be instant.
> Queries to the server are NOT required to be instant, so I'm wondering if this is the right approach.

@vinaybadgujar102 Care to comment?

vinaybadgujar102:
While checking for performance bottlenecks I noticed lot of unnecessary re renders. for ex: the header and bottom navbar re-render multiple times when navigating into a child goal even though they dont need to. Small things like this keep adding up and make the app feel laggy on low-end devices.

tanstack router and react router work almost the same, but tanstack adds strong typescript support for routes, params, and search params, which improves the DX a lot. It also integrates directly with React Query, which we already rely on for data fetching and caching.

The real performance issue isnt the router or the database, the DB queries are instant. The delay comes from React having to process and render components. On top of that the current app has a lot of nested, complex logic thats hard to follow and causes components to update more than they should. The component library we are using also limits fine grained control which makes optimization harder.

So I decided that I will be better to refactor some parts of the app with a clearer separation of concerns, simplifying the logic wherever possible. The goal is to make the code easier to understand and to ensure that future changes only require updating the parts that actually need it.

